PR #137, opened 2026-02-27
boilerplate tests

### LINKED ISSUE #86: [DEV] boilerplate code: write tests
**Location**

`/`

____

**Description**

Tests should be written for the boilerplate code.

100% coverage for both back-end and front-end since it's the boilerplate code.

Unit and integration tests are needed. (UI/E2E tests can wait)

(Note: I did not make this part of #84 because otherwise the team will have to wait even longer before finally starting to code, this is the ONLY exception)

____

**Sources**

noahvsb:
**Issue(s)**

closes #86 

____

**Description**

Wrote tests for the boilerplate code. I have achieved 100% coverage on frontend and 98% on backend.

For the backend [this line](https://github.com/SELab-2/viernulvier-1/blob/bd5123830006f46b434da96569f3a3f9f54b0ce4/backend/src/server.ts#L71) was impossible to test.

An idea could be to put that line in a `main.ts` file and run that in the docker instead of `server.ts`; then we would exclude `main.ts` from the coverage report and we'd achieve 100% coverage for backend.

____

**Sources**


backend/src/server.ts
```diff
@@ -1,40 +1,57 @@
-import Fastify from "fastify";
+import Fastify, { type FastifyInstance } from "fastify";
 import dbPlugin from "./db/postgres.js";
 import productionRoutes from "./routes/productions.js";
 
 import "dotenv/config";
 
-const isDebug = process.env["DEBUG"]?.toLowerCase() === "true";
-
-export const server = Fastify({
```
noahvsb, line 9: This export is useless and it makes the tests fail. If in the future the export is needed, then review your own code, because it shouldn't be necessary.

rundev.sh
noahvsb, line 1: Just a `chmod +x rundev.sh`. I thought it might be nice to push this as well so we wouldn't need to chmod it each time we pull.

backend/test/server.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    });
+  });
+
+  describe("start()", () => {
+    beforeEach(() => {
+      vi.resetModules();
+    });
+
+    test("starts on default port 3000 when BACKEND_PORT is unset", async () => {
+      const { start } = await import("@/server.js");
```
Ikerstreamer: Why are we using a mix of both await import as well as ESM module based imports?

backend/test/server.test.ts
```diff
@@ -0,0 +1,84 @@
+import { describe, test, expect, vi, beforeEach } from "vitest";
+
+vi.mock("@/db/postgres.js", () => ({
+  default: vi.fn(async (server: FastifyInstance) => {
+    server.pg = { query: vi.fn() } as any;
+  }),
+}));
+
+import { buildServer } from "@/server.js";
```
Ikerstreamer: Why aren't all module wide imports at the top of the file?

backend/src/server.ts
Ikerstreamer: I mainly wrote it like this to start because the export existed before the buildServer method which is a way nicer apporach.

rundev.sh
Ikerstreamer: This makes a lot of sense but I assumed I had already done this.

backend/src/server.ts
Simon-VdB, line 1: don't quite understand why that export caused problems, but I don't need to i think. It works the same way now, just not with that export. So I trust it.

frontend/test/components/HelloWorld.test.ts
```diff
@@ -0,0 +1,25 @@
+import { mount } from "@vue/test-utils";
+import { describe, expect, test } from "vitest";
+import App from "@/App.vue";
+
```
Ikerstreamer, line 4: These tests aren't that useful as we will probs be removing this standard index page generated by vue for a real homepage.

Ikerstreamer (changes requested):
Most of this looks pretty standard I did leave a few comments on some code quality stuff but for the most part it looks good.

Simon-VdB (dismissed):
Ran the tests and coverage, and it works and is indeed 98 and 100

frontend/test/components/HelloWorld.test.ts
noahvsb: True, but it can just serve as an example.

lucasodb:
Backend test for port 3000 fails, not immediately clear why.

noahvsb:
> Backend test for port 3000 fails, not immediately clear why.

I also don't understand, I recreated it by doing the following:

``` sh
pnpm run build-all
pnpm run coverage-all
```

But I was only able to recreate it once, every time after that the tests succeeded.

noahvsb:
> > Backend test for port 3000 fails, not immediately clear why.
> 
> I also don't understand, I recreated it by doing the following:
> 
> ```shell
> pnpm run build-all
> pnpm run coverage-all
> ```
> 
> But I was only able to recreate it once, every time after that the tests succeeded.

Update: I found the error, fixing it is pretty hard though, since it's very sporadic.

backend/src/server.ts
noahvsb: The global server variable messed with the server variable in my tests. In hindsight I could've maybe just renamed it in my tests to fix it, but the global variable is not necessary and it's a bad practice to have global variables.

backend/test/server.test.ts
noahvsb: Because of the `vi.resetModules()`, but I went for a different approach now that doesn't require the resetting of modules.

backend/test/server.test.ts
noahvsb: I thought this was necessary for the mock, but it wasn't. Fixed now.

noahvsb:
> Update: I found the error, fixing it is pretty hard though, since it's very sporadic

Update 2: fixed it

lucasodb (approved):
All tests now pass, with expected coverage.

Ikerstreamer (approved):
The tests do work and code quality looks up[ to par I suggest we merge this and then direct after it we merge #150. Or even just #150 as I did try and keep the branch up to date such that there would be 0 merge conflicts.

PR #140, opened 2026-02-28
added an admin model

### LINKED ISSUE #136: [DEV] admin object: make model
**Location**

`/shared/src/types/`

____

**Description**

Make a model for the admin information. Simple object, without any password
____

**Sources**

Build a model according to the database table for admin / the [EER](https://github.com/SELab-2/viernulvier-1/blob/staging/DOCS/database-schema.svg)

noahvsb:
**Issue(s)**

closes #136 

____

**Description**

Adds an admin model according to the [EER](https://github.com/SELab-2/viernulvier-1/blob/staging/DOCS/database-schema.svg).

____

**Sources**


willemDC-Info (dismissed):
Looks good to me. All reqiured fields are made and appear to be working as intended in the EER.

noahvsb:
I added a `MetadataSchema` so other `Schema`'s can extend it. This way any changes you make in the metadata, will only have to be made in `MetadataSchema`, not in each `Schema`.

Simon-VdB (dismissed):
Great, let's merge this.

shared/src/types/admin.ts
```diff
@@ -0,0 +1,15 @@
+import z from "zod"
+
+export const AdminSchema: z.ZodObject<any> = z.object({
+  // metadata (only time this is hardcoded)
```
Ikerstreamer: I would suggest adding to this comment that the proper way to do it for other objects is. using a spread with 

```
...MetadataSchema.shape
```

I prefer this over using zod's in build extend as extend can get messy if u want to do composition instead of inheritance, and it also beter follows our EER diagram as well as the postgress standard for using partial tables.

shared/src/index.ts
```diff
@@ -1 +1 @@
-export * from "./types/production.js";
\ No newline at end of file
+export * from "./types/index.js";
```
Ikerstreamer, line 1: Doing one index per sub module makes a lot of sense.

shared/src/types/admin.ts
Ikerstreamer: You could also add to the comment that this is to prevent an import loop.

Ikerstreamer (commented):
I would suggest also adding the metadata to the production module such that it shows the intention.

Ikerstreamer (changes requested):
Fixing the listed comment as well as maybe even adding the metadata into the productionSchema type would be grand.

noahvsb:
> Fixing the listed comment as well as maybe even adding the metadata into the productionSchema type would be grand.

The metadata is added to ProductionSchema in #142. I think it's better I don't touch it so we don't merging conflicts. If people want to see the intentions I suggest they check out #142.

Simon-VdB (dismissed):

willemDC-Info (dismissed):
Looks good to me. I'll make changes to my own branch to account fot the metadata field so we can import it to all other types if needed

Ikerstreamer:
Could u also add .describe("is this") to all zod param types as well as to the entire object. As an example

```ts

export const ProductionSchema = z
  .object({
    id: z.number().describe("id for the production."),
    title: z.string().describe("title for the given production."),
  })
  .describe("Production");
  ```

**Comment by noahvsb** [comment]
> Could u also add .describe("is this") to all zod param types as well as to the entire object. As an example

This is out of scope: #95

File: shared/src/types/admin.ts [hunk 3]
```diff
@@ -0,0 +1,17 @@
...
+export const AdminSchema: z.ZodObject<any> = z.object({
+  // METADATA
+  // only time this is hardcoded to avoid infinite import loops
+  // normally you would use ...MetadataSchema.shape instead of hardcoding
+  created_by: z.lazy(() => AdminSchema),
+  created_at: z.date(),
+  updated_by: z.lazy(() => AdminSchema),
+  updated_at: z.date(),
+
+  id: z.number(),
```
Ikerstreamer: Even if we don't wanna add desciptions I would still suggest atleast making the types as fixed as possible as this does get checked when u do `AdminSchema.safeParse(JSON_OBJECT)`

```suggestion
  id: z.int().positive(),
```

shared/src/types/admin.ts
```diff
@@ -0,0 +1,17 @@
...
+  username: z.string(),
```
Ikerstreamer: Add the max length of the string for anything that is a varchar in the backend.

```suggestion
  username: z.string().max(32),
```

shared/src/types/admin.ts
```diff
@@ -0,0 +1,17 @@
...
+  profile_picture: z.string().nullable(),
```
Ikerstreamer: There is a special .url() string type within zod that actually checks the string is the right format. And once again max length.
```suggestion
  profile_picture: z.url().max(2048).nullable(),
```

shared/src/types/admin.ts
```diff
@@ -0,0 +1,17 @@
...
+});
```
Ikerstreamer: We also seem to be missing a password field.

Ikerstreamer (changes requested):

shared/src/types/admin.ts
noahvsb: In the online meeting last friday, I thought we decided to only have the password in the database, front-end shouldn't have access to this.

Simon-VdB (dismissed):

shared/src/types/admin.ts
Simon-VdB, line 1: Don't know why this is supposed to be better, but alright, if you and Iker know that, that's fine by me

shared/src/types/metadata.ts
Simon-VdB, line 1: same here

Simon-VdB (dismissed):

shared/src/types/admin.ts
```diff
@@ -0,0 +1,18 @@
...
+  id: primaryKey(),
+  username: z.string().max(32),
+  profile_picture: z.url().max(2048).nullable(),
+};
+
+type AdminSchemaType = z.ZodObject<typeof AdminBase> & {
+  withMeta: () => z.ZodObject<typeof AdminBase & typeof MetadataShape>;
+};
+
+export const AdminSchema: AdminSchemaType = createSchema(AdminBase);
```
Simon-VdB, line 15: sheesh, oneliner 🥇

Simon-VdB (dismissed):

Ikerstreamer (approved):

Issue #12, opened 2026-02-21
[DEV] implement EER-diagram in SQL


noahvsb:
**Location**

`/backend/migrations/`

____

**Description**

Implement the full EER-diagram with corresponding types in SQL in the /backend/migrations/ folder.

____

**Sources**

(outdated)
<img width="1582" height="1239" alt="Image" src="https://github.com/user-attachments/assets/9e17051e-6a8d-4303-a106-8cb808b38822" />

PR #145, opened 2026-02-28
Feat/database migrations 2

Acher0s:
**Issue(s)**

#12 

____

**Description**

added migrations for rest of tables in EER

____

**Sources**

#87


backend/migrations/007.do.create-production-custom-field-logic-tables.sql
```diff
@@ -0,0 +1,58 @@
...
+
+  -- metadata
+  created_by  INT             REFERENCES admin(id) ON DELETE SET NULL,
+  created_at  TIMESTAMPTZ     NOT NULL DEFAULT NOW(),
+  updated_by  INT             REFERENCES admin(id) ON DELETE SET NULL,
+  updated_at  TIMESTAMPTZ     NOT NULL DEFAULT NOW()
+);
+
+-- ============================================================
+-- PRODUCTION CUSTOM FIELD (The EAV Table)
```
Simon-VdB, line 35: what does EAV stand for?

backend/migrations/007.do.create-production-custom-field-logic-tables.sql
```diff
@@ -0,0 +1,58 @@
...
+-- ============================================================
+CREATE TABLE IF NOT EXISTS production_custom_field (
+  field_id        INT             REFERENCES custom_production_field_definition(id) ON DELETE CASCADE,
```
Simon-VdB: isn't defenition_id better? I'm just thinking out loud. because you call it field_id, but it's a foreign key from the field_defenition...

noahvsb: I agree.

backend/migrations/007.do.create-production-custom-field-logic-tables.sql
noahvsb: https://en.wikipedia.org/wiki/Entity%E2%80%93attribute%E2%80%93value_model

PR #150, opened 2026-03-01
Add coverage actions

### LINKED ISSUE #89: [WORKFLOW] Set a threshold for coverage tests, implement into actions
**Idea**
In #88 I added basic coverage testing as vitest, our test framework, supports it out of the box. However, I did not fully configure it with a coverage threshold. [Here](https://vitest.dev/config/coverage#coverage-thresholds) you can find the docs for configuring vitest coverage. 

By default, there is no threshold, and thus the test always succeeds given that the unit test also succeeds, as that is a prerequisite.

I would suggest we set the coverage to atleast 75%. It could even be higher, given that I think we should only include files in ./src when doing our coverage test.

If we set up a threshold, I suggest we then also include how this is into our PR actions run by github as having good tests seems important. 

We might want to only do this for staging/main and not all dev branches as it is quite a bit more intense than running the tests without the coverage.

Ikerstreamer:
**Issue(s)**

closes #89

implements changes from #137 this was needed to get the coverage up to the desired levels.

____

**Description**

Setup coverage thresholds in the config files and make it such that github PRs to staging/main check coverage before permitting a merge.


backend/vitest.config.ts
```diff
@@ -13,6 +13,20 @@ export default defineConfig({
...
       include: ["src/**/*.{ts,tsx}"],
+      thresholds: {
+        "src/*/**/*.{ts,tsx}": {
+          statements: 97.5,
+          functions: 97.5,
+          branches: 97.5,
+          lines: 97.5,
+        },
+        perFile: true,
+        "src/server.ts": {
```
noahvsb: You can remove this now that server.ts has 100% coverage on #137.

frontend/vite.config.ts
```diff
@@ -29,6 +29,15 @@ export default defineConfig({
     coverage: {
       provider: "v8",
       include: ["src/**/*.{ts,tsx}"],
+      thresholds: {
+        "src/*/**/*.{ts,tsx}": {
+          statements: 97.5,
+          functions: 97.5,
+          branches: 97.5,
+          lines: 97.5,
```
noahvsb: We decided on 80% for the frontend.

noahvsb (changes requested):

backend/src/server.ts
Simon-VdB, line 1: I think Noah already did this, no? He explained that export needed to go away to be able to succeed at all tests

frontend/vite.config.ts
Simon-VdB, line 1: Like Noah said: see meeting notes of last wednesday

Simon-VdB (dismissed):
Looks good, but it seems as if you and Noah are doing some "dubbel werk"? Or did one of you pull from the other. It's just that I've reviewed a very similar pr like this.

backend/src/server.ts
Ikerstreamer: If u read my description it states that I implemented Noah pr into mine. As that would allow the coverage to actually succeed.

noahvsb (dismissed):
Looks good now, ready to merge after #137 has been merged 👍

noahvsb (approved):

msth5 (approved):
Looking good, not many changes, and even fewer actual changes, but those made are looking fine

PR #153, opened 2026-03-04
fixed path mistake

Acher0s:
**Issue(s)**

N/A

____

**Description**

path in dockerfile was wrong, now its correct (determined after local testing)

____

**Sources**

N/A

### DIFF
--- a/backend/Dockerfile
+++ b/backend/Dockerfile
@@ -50,5 +50,5 @@ WORKDIR /app
 COPY --from=build /app/backend/dist ./dist
 COPY --from=build /app/node_modules ./node_modules
 
-EXPOSE 3000
+EXPOSE $BACKEND_PORT
 CMD ["node", "dist/backend/src/index.js"]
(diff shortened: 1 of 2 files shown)


backend/Dockerfile
```diff
@@ -50,4 +50,5 @@ WORKDIR /app
 COPY --from=build /app/backend/dist ./dist
 COPY --from=build /app/node_modules ./node_modules
 
-CMD ["node", "dist/server.js"]
\ No newline at end of file
+EXPOSE 3000
```
Ikerstreamer: I would suggest using the env variable.

Ikerstreamer (dismissed):

Ikerstreamer:
I would also suggest updating the vite.config file for the frontend if we are planning to use that as proxy.

Ikerstreamer (dismissed):

Ikerstreamer (approved):

PR #154, opened 2026-03-04
Fix/frontend dockerfile port

Acher0s:
**Issue(s)**

N/A

____

**Description**
frontend dockerfile now uses env to detirmine exposed port

____

**Sources**

N/A


Ikerstreamer (commented):
I would suggest merging this with #153.

PR #155, opened 2026-03-04
fix expose port to 80

Acher0s:
**Issue(s)**

N/A
____

**Description**

fix expose port to 80, it has to be hardcoded somewhere since the build doesn't happen on the hosting server. can still be mapped as desired in docker compose so there's no loss of flexibility.

____

**Sources**

N/A


noahvsb:
Why is only 1 review required?

noahvsb (approved):

PR #156, opened 2026-03-04
Feat/image model

### LINKED ISSUE #143: [DEV] image object: make model
**Location**

`/shared/src/types/`

____

**Description**

Make a model for the image and crop objects
____

**Sources**

Build a model according to the database table for image and crop / the [EER](https://github.com/SELab-2/viernulvier-1/blob/staging/DOCS/database-schema.svg)

willemDC-Info:
**Issue(s)**

____

**Description**

____

**Sources**


willemDC-Info:
Made some adjustments to the image and crop type based on the changes that where made to other types.
(Mainly adding propper input validation to all fields with accordance to the EER)

noahvsb (dismissed):
Consistent with other models 👍

shared/src/types/image.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export const ImageSchema: SchemaWithMeta<any> = createSchema({
+  id: primaryKey(),
+
+  production_id: foreignKey(() => ProductionSchema),
+
+  res: z
+    .string()
+    .max(16)
+    .min(1)
```
msth5: is this not unnecessary with your refine later one?

msth5 (dismissed):
In migrations you can see that the actual url can be 2048 long
on https://github.com/SELab-2/viernulvier-1/blob/8ef7ec6c4907acba8955c04c499e081a3136801a/backend/migrations/005.do.create-image-tables.sql#L18

shared/src/types/image.ts
willemDC-Info: Yeah I think I can remove the min because of the trim later that checks greater then 0.

willemDC-Info (commented):
All requested changes have been made to image.ts and crop.ts.

- I removed unnecessary checks for minimum length in image.ts
- I increased the size of the allowed crop URL from 128 to 2048

msth5 (approved):
Suggestions implemented, I approve

PR #158, opened 2026-03-04
Hotfix for broken sql migration files.

Ikerstreamer:
**Issue(s)**

N/A

____

**Description**

The migration files currently on staging contain sql with invalid syntax. This quick hotfix makes sure we have working sql.

____

**Sources**

Check the dir backend/migrations.
I also changed how the migrate.ts file works so we can actually undo migrations now.


migrate-db.sh
noahvsb, line 1: This runs the same command twice for me.

It seems to work, I get a list of migrations, but it ends with `Migration completed: undefined`.

noahvsb (changes requested):

noahvsb (approved):

Issue #132, opened 2026-02-27
[DEV] event object: make model


msth5:
**Location**

`/shared/src/types/`

____

**Description**

Make a model for the event object
____

**Sources**

Build a model according to the database table for event / the [EER](https://github.com/SELab-2/viernulvier-1/blob/staging/DOCS/database-schema.svg)

PR #142, opened 2026-02-28
Feat/production model

noahvsb:
**Issue(s)**

closes #131

____

**Description**

Added a production model according to the [EER](https://github.com/SELab-2/viernulvier-1/blob/staging/DOCS/database-schema.svg).

____

**Sources**

PR #163, opened 2026-03-05
Feat/event price model

msth5:
**Issue(s)**

#132 Addendum

____

**Description**

Added a price model and added the pricing to the events

____

**Sources**

n/a


shared/src/types/event.ts
```diff
@@ -12,6 +13,7 @@ export const EventSchema = createSchema({
     doors_at: z.date(),
     vendor_id: z.int().nonnegative(),
     info: languageMap,
+    pricing: z.array(foreignKey(() => PriceSchema)),
```
noahvsb:
```suggestion
    price: z.array(foreignKey(() => PriceSchema)),
```

noahvsb (changes requested):
Rename to `EventPriceSchema` to be consistent with the SQL. Since it's only for events, maybe also move the code to `event.ts`.

lucasodb (approved):
This schema makes sense to me, looks good.

shared/src/types/event.ts
msth5: Done while merging price and event file

noahvsb (changes requested):
You forgot to rename `PriceSchema` to `EventPriceSchema` for consistency with the migrations.

noahvsb:
I'll just do the small change in #142

PR #166, opened 2026-03-06
fix schema typing

noahvsb:
**Issue(s)**

____

**Description**

The schema typings didn't work, I couldn't use `AdminSchema.pick()`. So I fixed it for all schema's.

____

**Sources**


msth5 (approved):
As long as either event or production has type declared specifically for circulair referencing

Issue #35, opened 2026-02-21
[DEV] events endpoint: delete event


noahvsb:
**Location**

`/backend/src/routes/`

____

**Description**

Implement a function to delete an event.

Handle `DELETE /api/v1/events/{id}`.

____

**Sources**

Issue #34, opened 2026-02-21
[DEV] events endpoint: bulk edit events


noahvsb:
**Location**

`/backend/src/routes/`

____

**Description**

Implement a bulk edit of events.

Not sure what HTTP request method we will use. There are 2 options:
- do a PUT or PATCH for each event that is edited (simple, but slower)
- do a PUT or PATCH to `/api/v1/events/bulk` with a body that has a set of id's and the fields that need to be edited (more complex, but faster)

Also something to think about is how errors will be handled. There are 2 options for this as well:
- if one fails, fail all
- if one fails, only this fails (seems better at first glance, but it's less user friendly, the user will have to look for that one failed event to retry)

____

**Sources**

Issue #33, opened 2026-02-21
[DEV] events endpoint: edit event


noahvsb:
**Location**

`/backend/src/routes/`

____

**Description**

Create a function to edit an event.

Handle `PUT /api/v1/events/{id}` and `PATCH /api/v1/events/{id}` with a json body of the editted fields.

____

**Sources**

Issue #32, opened 2026-02-21
[DEV] events endpoint: create event


noahvsb:
**Location**

`/backend/src/routes/`

____

**Description**

Implement a function to create an event linked to a production.

Handle `POST /api/v1/events` with a json body of the event.

____

**Sources**

PR #167, opened 2026-03-06
Feat/event endpoints

msth5:
**Issue(s)**

Closing #30, #31, #32, #33, #34 & #35

____

**Description**

Implementing fetch methods and mapping them on get requests for event objects

____

**Sources**

n/a


noahvsb:
@msth5 I think it's best to have 1 big PR for the whole event endpoint (and any endpoint in general). Because it's better to see the whole picture when reviewing and otherwise it will give us too many small PR's and will be a headache to keep up with.

PR #168, opened 2026-03-06
authentication endpoint + authorization

### LINKED ISSUE #46: [DEV] API endpoints: authentication
**Location**

`/backend/src/routes/auth/`

____

**Description**

Implement the authentication endpoint of the API

`/api/v1/auth/` (?)

____

**Sources**

### LINKED ISSUE #78: [DEV] authentication endpoint: fetch admin
**Location**

`/backend/src/routes/auth/`

____

**Description**

Implement a function to fetch an admin.

Handle `GET /api/v1/auth/{id}`.

____

**Sources**

### LINKED ISSUE #79: [DEV] authentication endpoint: create admin
**Location**

`/backend/src/routes/auth/`

____

**Description**

Implement a function to create an admin.

Handle `POST /api/v1/auth/` with a json body of the admin.

____

**Sources**

### LINKED ISSUE #80: [DEV] authentication endpoint: edit admin
**Location**

`/backend/src/routes/authg/`

____

**Description**

Create a function to edit an admin.

Handle `PUT /api/v1/auth/{id}` and `PATCH /api/v1/auth/{id}` with a json body of the editted fields.

____

**Sources**

### LINKED ISSUE #81: [DEV] authentication endpoint: delete admin
**Location**

`/backend/src/routes/auth/`

____

**Description**

Implement a function to delete an admin.

Handle `DELETE /api/v1/aiuth/{id}`.

____

**Sources**

### LINKED ISSUE #82: [DEV] authentication endpoint: admin login
**Location**

`/backend/src/routes/auth/`

____

**Description**

Implement a function to login as admin.

Handle `POST /api/v1/auth/login/` with a json body with username and password.
____

**Sources**

### LINKED ISSUE #83: [DEV] authorisation: allow only admins to do non-fetch and admin and metadata fetch API calls
**Location**

`/backend/src/`

____

**Description**

Only admins who are logged in should be allowed to do all of the non-fetch API calls (create, edit, delete) and the admin and metadata fetch API call.

____

**Sources**

noahvsb:
**Issue(s)**

closes #78, #79, #80, #81, #82, #83

____

**Description**

Implemented the authentication endpoint (without tests).

Haven't implemented authorization yet.

____

**Sources**


noahvsb:
accidentally used main as target

noahvsb:
Staging validation is failing because the feat/endpoint-helpers branch still needs to implement tests.

backend/src/routes/auth/handlers/login.ts
```diff
@@ -0,0 +1,46 @@
...
+import { z } from "zod";
+import { comparePassword } from "./hash.js";
+
+const LoginBodySchema = AdminSchema.pick({ username: true }).extend({
+  password: z.string(),
+});
+
+/**
+ * Authenticates an admin by username and password.
+ * Returns a placeholder session token until proper auth is implemented.
```
Simon-VdB: It's not a placeholder anymore.

backend/src/routes/auth/handlers/login.ts
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * @param server - The Fastify instance, used for database access and logging.
+ * @param request - The Fastify request, expected to contain `username` and `password` in its body.
+ * @returns An object containing a placeholder session token.
```
Simon-VdB: same here, not a placeholder anymore right?

backend/scripts/generate-secret.ts
```diff
@@ -0,0 +1,4 @@
+import crypto from "crypto";
+
+const secret = crypto.randomBytes(32).toString("hex");
+console.log(`JWT_SECRET=${secret}`);
```
Simon-VdB, line 4: why are we logging this?

backend/src/plugins/authorize.ts
```diff
@@ -0,0 +1,25 @@
...
+import type { FastifyInstance, FastifyRequest, FastifyReply } from "fastify";
+
+/**
+ * Decorates the Fastify instance with an `authorize` hook that verifies the JWT session cookie.
+ * Attach this as a `preHandler` on routes that require authentication.
+ * Returns a `401 Unauthorized` response if the token is missing or invalid.
+ *
+ * @param server - The Fastify instance to decorate.
+ */
+export default fp(function authorizePlugin(server: FastifyInstance) {
```
Simon-VdB, line 11: so here you add the authorize function to the server...

backend/src/plugins/authorize.ts
```diff
@@ -0,0 +1,25 @@
...
+  server.decorate("authorize", async function (request: FastifyRequest, reply: FastifyReply) {
+    try {
+      await request.jwtVerify();
+    } catch {
+      reply.status(401).send({ error: "Unauthorized" });
+    }
+  });
+});
+
+declare module "fastify" {
```
Simon-VdB, line 21: and here you just tell typescript that server.authorize will exist?

Simon-VdB: so in js this wouldn't be necessary?

backend/src/plugins/jwt.ts
```diff
@@ -0,0 +1,23 @@
...
+ * Registers JWT and cookie plugins on the Fastify instance.
+ * The JWT secret is read from the `JWT_SECRET` environment variable.
+ * Sessions are stored in an `httpOnly` cookie named `session`.
+ *
+ * @param server - The Fastify instance to register the plugins on.
+ */
+export default fp(async function jwtPlugin(server: FastifyInstance) {
+  await server.register(fastifyCookie);
+
+  await server.register(fastifyJwt, {
```
Simon-VdB, line 22: So this is the servers "private key" for validating the tokens sent by users in their requests either in the authorization header or as a cookie?

backend/src/routes/auth/handlers/edit.ts
```diff
@@ -0,0 +1,49 @@
...
+  if (body.password !== undefined) {
+    fields.push(`password = $${i++}`);
+    values.push(await hashPassword(body.password));
+  }
+
+  fields.push(`updated_by = $${i++}`, `updated_at = $${i++}`);
+  values.push(admin, current_time, id);
+
+  const result = await server.pg.query<Admin>(
+    `UPDATE admin SET ${fields.join(", ")} WHERE id = $${i}
```
Simon-VdB, line 42:
```suggestion
    `UPDATE admin SET ${fields.join(", ")} WHERE id = $${i} // secure way of putting user input in database query
```

backend/scripts/generate-secret.ts
Simon-VdB: is this just for example purposes?

backend/src/routes/auth/handlers/login.ts
Simon-VdB:
```suggestion
 * @returns { success: true } if cookie is successfully set up with the session token expiring in 24h.
```

backend/src/routes/auth/handlers/login.ts
Simon-VdB, line 1: I have a few more questions about this: what if I'm curling the route from my terminal? Then I don't have cookies, so then I would have to use the Authorization header, but I don't see this in your code yet.

backend/src/routes/auth/auth.ts
```diff
@@ -0,0 +1,31 @@
...
+ * - `PUT /api/v1/auth/:id` — replace an existing admin's data. 🔒
+ * - `PATCH /api/v1/auth/:id` — partially update an existing admin. 🔒
+ * - `DELETE /api/v1/auth/:id` — delete an admin by ID. 🔒
+ * - `POST /api/v1/auth/login` — authenticate an admin and set a session cookie.
+ * - `POST /api/v1/auth/logout` — clear the session cookie.
+ *
+ * @param server - The Fastify instance to register routes on.
+ */
+export default function authRoutes(server: FastifyInstance) {
+  const protect = { preHandler: [server.authorize] };
```
Simon-VdB, line 21: very clean, well done, easy to use and very easy to read the code like this

backend/src/routes/auth/auth.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  server.get("/api/v1/auth/:id", protect, replyHandler(server, fetchAdmin));
+  server.get("/api/v1/auth/:id/meta", protect, replyHandler(server, fetchAdminWithMeta));
+  server.post("/api/v1/auth", protect, replyHandler(server, createAdmin));
+  server.put("/api/v1/auth/:id", protect, replyHandler(server, replaceAdmin));
+  server.patch("/api/v1/auth/:id", protect, replyHandler(server, editAdmin));
+  server.delete("/api/v1/auth/:id", protect, replyHandler(server, deleteAdmin));
+  server.post("/api/v1/auth/login", replyHandler(server, login));
```
Simon-VdB, line 29: still trying to figure out how the endpoint knows if it should look for cookie, or for authorization header

backend/src/routes/helpers.ts
```diff
@@ -0,0 +1,121 @@
...
+
+/**
+ * Wraps a handler function and sends a 404 response if the result is null.
+ * Catches HttpErrors and forwards their status code and message.
+ *
+ * @param server - The Fastify instance, used for route registration.
+ * @param handler - The handler function to wrap.
+ * @returns A Fastify route handler.
+ */
+export function replyHandler<T>(
```
Simon-VdB: This is very nice, because it forces you to write handlers with a specific interface, but maybe in the future we will be confronted with a situation where in we cannot follow this interface? Or can you give a good reason why we can never run into such situations?

backend/src/plugins/jwt.ts
```diff
@@ -0,0 +1,23 @@
...
+    secret: process.env["JWT_SECRET"]!,
```
Simon-VdB: The ! non-null assertion bypasses TypeScript safety. If the env var isn't set, Fastify will silently get undefined as the JWT secret and likely crash at runtime on first use rather than at startup. Better to validate early and throw a clear error:

```suggestion
    const secret = process.env["JWT_SECRET"];
    if (!secret) throw new Error("JWT_SECRET environment variable is not set");
```

backend/src/routes/auth/handlers/login.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+export async function login(server: FastifyInstance, request: FastifyRequest, reply: FastifyReply) {
+  const { username, password } = parse(server, LoginBodySchema, request.body);
+
+  const result = await server.pg.query(
+    `SELECT id, password FROM admin WHERE username = $1`,
+    [username]
+  );
+
+  if (result.rows.length === 0) throw new HttpError(401, "Invalid credentials");
```
Simon-VdB: When the username doesn't exist, the response returns instantly (no bcrypt). When it does exist but the password is wrong, bcrypt takes ~100ms. This timing difference lets an attacker enumerate valid usernames. The fix is to always run comparePassword against a dummy hash when no user is found.

backend/src/routes/auth/handlers/edit.ts
Simon-VdB, line 1: In editAdmin, if the request body is {}, fields only ends up with updated_by and updated_at, resulting in a no-op query that still returns 200. Consider adding a check that at least one meaningful field is provided.

Simon-VdB (commented):
Very well designed and implemented. 
- All tests succeed.
- !!! I'm still not able to run the docker containers: i get this error:
/app/shared/src/types/metadata.ts:23
viernulvier-backend   |     value: () => base.extend(MetadataShape),
viernulvier-backend   |            ^
viernulvier-backend   | 
viernulvier-backend   | TypeError: __name is not a function

Earlier I got an error saying that @fastify/jwt was not found. But that seems to be gone now.

I have a few thoughts I would like to share, maybe they are clearer for you:

1) To me it looks like you only work with cookies, but what if I want to send a request from my terminal or let's say postman, I would have to use the authorization header instead of the cookies, right? 

2) Same for your tests, you only test methods with cookies.

3) Maybe the authorization header logic is yet to be implemented? Maybe it is standard in fastify routes? Maybe I'm missing something else?

Some things to consider:
1) The CRUD operations on /auth are really on admin related things, so maybe we should use /admin for those and leave /auth as an endpoint purely for login and logout? Because that one would be public.

2) login.test.ts imports hashPassword from the handlers index to set up mock data. This leaks an implementation detail into the public surface of the handlers module. A small utility like that is better imported directly from hash.ts in the test, or the test could use a pre-hashed constant.

backend/scripts/generate-secret.ts
noahvsb: You'll have to generate your own secret for your dev environment. This script generates one and logs the output, so you can copy paste it from your terminal into your `.env`.

backend/src/plugins/authorize.ts
noahvsb: I'm not sure, `declare module` is typescript syntax, but perhaps you also need to do something in javascript. I'm not too familiar with js.

backend/src/plugins/jwt.ts
noahvsb: Yes

backend/src/routes/auth/handlers/edit.ts
noahvsb: I'm not sure it's a secure way per se, it's just because not all fields will be changed. Otherwise you would need a different query for each subset of changed fields.

backend/src/routes/auth/handlers/login.ts
noahvsb: I don't think we support authorization headers, only cookies. But I think using curl you're also able to provide a cookie. Maybe jwt supports authorization headers automatically, I'm not sure and this public wifi is horrible, so I don't want to research it.

backend/src/routes/helpers.ts
noahvsb: We can always follow the interface, in some cases we just won't use all of the variables.
If we need more variables than the interface offers, you may need to rethink if that's really necessary and if you're convinced, we can always update the interface.

backend/src/plugins/jwt.ts
noahvsb: I thought when using ? it will silently set it as undefined and when using ! it would throw an error.

backend/src/routes/auth/handlers/edit.ts
noahvsb: If people really want to use an edit and edit nothing, we shouldn't stop them from pursuing their dreams.

The [official specs](https://www.rfc-editor.org/rfc/rfc5789.html) don't state that empty bodies aren't allowed.

willemDC-Info (dismissed):
All code looks good to me. Main thing holding this pr back is the testcoverage for helper.ts is at 95%.

backend/scripts/generate-secret.ts
Simon-VdB: makes sense now, thanks.

backend/src/routes/auth/handlers/edit.ts
Simon-VdB: I thought doing it that way you are protected against sql injection..

backend/src/routes/auth/handlers/login.ts
Simon-VdB: Maybe we should ask our "colleagues" this? I personally think authorization headers would be useful.

backend/src/routes/auth/handlers/edit.ts
Simon-VdB: But then metadata would be junk, right?

backend/src/plugins/jwt.ts
Simon-VdB: The ! is TypeScript's non-null assertion operator — it is a compile-time only hint to the TypeScript compiler that says "trust me, this won't be null or undefined." TypeScript uses it to stop showing type errors. When the code gets compiled to JavaScript, the ! is completely stripped out and leaves no trace whatsoever.

So at runtime, this:
secret: process.env["JWT_SECRET"]!

is identical to this:
secret: process.env["JWT_SECRET"]

If the env var isn't set, you get undefined silently passed as the JWT secret. No error, no warning.

backend/src/routes/auth/handlers/edit.ts
noahvsb: if they try to use  sql injection, z.parse() will fail

backend/src/routes/auth/handlers/login.ts
noahvsb: true, but maybe for milestone 2 or 3

backend/src/plugins/jwt.ts
noahvsb: I got confused with dart

backend/src/routes/auth/handlers/edit.ts
noahvsb: No, because you editted the metadata

backend/src/routes/auth/handlers/login.ts
noahvsb: I made an issue for this: #177

backend/src/routes/auth/handlers/replace.ts
```diff
@@ -0,0 +1,47 @@
...
+import { hashPassword } from "./hash.js";
+
+const ReplaceAdminBodySchema = AdminSchema.pick({ username: true }).extend({
+  password: z.string().min(8).max(72),
+});
+
+const replaceAdminById = (server: FastifyInstance) =>
+  buildQuery(
+    server,
+    `UPDATE admin SET username = $1, password = $2, updated_by = $3, updated_at = $4
```
msth5, line 15: Update only update if exists, however you use this as your full functionality for your put endpoint and that also needs to add objects when not present in the db

backend/src/routes/auth/handlers/hash.ts
```diff
@@ -0,0 +1,24 @@
...
+}
+
+/**
+ * Compares a plain text password against a bcrypt hash.
+ *
+ * @param password - The plain text password to verify.
+ * @param hash - The bcrypt hash to compare against.
+ * @returns `true` if the password matches the hash, `false` otherwise.
+ */
+export async function comparePassword(password: string, hash: string): Promise<boolean> {
```
msth5, line 22: When are we gonna have to compare plain text passwords with hashes? shouldn't passwords always get hashed on client side?

msth5 (changes requested):

backend/src/routes/helpers.ts
```diff
@@ -0,0 +1,345 @@
...
+ * @param request - The Fastify request to extract params from.
+ * @param key - The parameter key to extract.
+ * @returns The parameter value as a string.
+ * @throws `Error` if the parameter is not present in the request.
+ */
+export function getParam(request: FastifyRequest, key: string): string {
+  // eslint-disable-next-line security/detect-object-injection
+  const value = (request.params as Record<string, string>)[key];
+  if (value === undefined)
+    throw new HttpError(400, `Missing route parameter: "${key}"`);
```
msth5:
```suggestion
    throw new HttpError(HttpClientError.BadRequest, `Missing route parameter: "${key}"`);
```
HTTP error code is hard coded when you have enum aviable

msth5: Iker forgot this one I think

.github/workflows/pr-dev.yml
```diff
@@ -17,7 +17,10 @@ jobs:
     runs-on: [self-hosted, linux, ARM64]
     strategy:
       matrix:
-        project: [frontend, backend]
+        project: [frontend, backend, shared]
+        include:
+          - project: shared
+            skip_tests: true
```
Simon-VdB, line 23: So we skip the shared tests?

backend/src/routes/auth/handlers/create.ts
```diff
@@ -1,33 +1,45 @@
...
+import { getMetadata, parseSchema, buildQuery } from "@/routes/helpers.js";
 import { z } from "zod";
 import { hashPassword } from "./hash.js";
 
 const CreateAdminBodySchema = AdminSchema.pick({ username: true }).extend({
   password: z.string().min(8).max(72),
 });
 
+const insertAdmin = (server: FastifyInstance) =>
+  buildQuery(
```
Simon-VdB, line 13: I don't get what was wrong with the previous way of building the query but alright, Iker will have an explanation I assume...

backend/src/routes/auth/handlers/edit.ts
```diff
@@ -39,11 +38,11 @@ export async function editAdmin(server: FastifyInstance, request: FastifyRequest
   fields.push(`updated_by = $${i++}`, `updated_at = $${i++}`);
   values.push(admin, current_time, id);
 
-  const result = await server.pg.query<Admin>(
+  const result = await server.pg.query(
```
Simon-VdB, line 41: what's the reason for not using the query builder here?

backend/src/routes/auth/handlers/login.ts
```diff
@@ -42,5 +54,5 @@ export async function login(server: FastifyInstance, request: FastifyRequest, re
     path: "/",
   });
 
-  return { success: true };
+  return { succes: true };
```
Simon-VdB: typo? : success

backend/test/server.test.ts
```diff
@@ -4,6 +4,7 @@ import type { FastifyInstance } from "fastify";
 
 vi.mock("@/plugin/postgres.js", () => ({
```
Simon-VdB: isn't it plugin**s** plural?

backend/src/routes/auth/handlers/login.ts
```diff
@@ -1,38 +1,50 @@
...
 
-  const valid = await comparePassword(password, result.rows[0].password);
+  // always compare hash to prevent a timing attack
+  const valid = await comparePassword(password, rows[0]?.password ?? DUMMY_HASH);
   if (!valid) throw new HttpError(401, "Invalid credentials");
 
   const token = server.jwt.sign(
-    { id: result.rows[0].id, username },
-    { expiresIn: "24h" }
+    { id: rows[0]!.id, username },
```
Simon-VdB, line 47: smelly

Simon-VdB (commented):
Looks good, tests succeed with 100% coverage. I can run the docker containers and access the auth route but ofc I'm unauthorized so i can't test it very well. 

I'm still having doubts about only using cookies, because our project is supposed to be api driven and by assuming only browsers will be used to view our api data, we are now putting importance on the frontend rather than the backend. Because with cookies, the ease of use in a browser is way better than with a curl request or via postman f.e.

backend/src/routes/auth/handlers/replace.ts
noahvsb: note: we decided in the discord we will not create a new object if the id doesn't exist. The [standard](https://www.rfc-editor.org/rfc/rfc2616#section-9.6) doesn't say it's necessary and we would like to have more controls over the id's that are in use.

backend/src/routes/auth/handlers/hash.ts
noahvsb: Leaving the hashing over to the back-end is a cleaner approach and we assume use of HTTPS so it's encrypted when traveling on the network..

noahvsb:
> I'm still having doubts about only using cookies, because our project is supposed to be api driven and by assuming only browsers will be used to view our api data, we are now putting importance on the frontend rather than the backend. Because with cookies, the ease of use in a browser is way better than with a curl request or via postman f.e.

I've come to agree with this. I made an issue, but I consider it out of scope for this PR: #177

.github/workflows/pr-dev.yml
noahvsb: Yes, because there aren't any.

backend/src/routes/auth/handlers/create.ts
noahvsb: Build query adds data validation to input and output.

backend/src/routes/auth/handlers/edit.ts
noahvsb: The way I assumed buildQuery is used is to build a constant query variable. But the query here isn't constant.

backend/src/routes/auth/handlers/login.ts
noahvsb: I now did `if (!valid || rows.length == 0) throw new HttpError(401, "Invalid credentials");` in the line above, so I can safely do the `!`.

Simon-VdB (dismissed):

willemDC-Info (dismissed):
All looks good to me and the hard coded part is out. I looks all good to me to merge it with staging.

noahvsb:
RIP approvals :pray:

backend/test/routes/auth/integration.test.ts
```diff
@@ -0,0 +1,260 @@
...
+          ? { rows: [{ id: row.id, password: row.password }], rowCount: 1 }
+          : { rows: [], rowCount: 0 }
+        );
+      }
+
+      // Fetch with meta: SELECT ... created_at ... WHERE id = $1
+      if (q.startsWith("SELECT") && q.includes("CREATED_AT")) {
+        const id = params[0] !== undefined ? Number(params[0]) : undefined;
+        const rows = id !== undefined
+          ? mockDb.filter((a) => a.id === id).map(({ password: _, ...rest }) => rest)
```
willemDC-Info, line 73: To make these errors go away you can use
/* eslint-disable no-unused-vars */
to remove them as they are no real errors but eslint complaining

noahvsb: I just decided to add a `...IgnorePattern: "^_"` like we did for the source code in the eslint config.

And for the object-sink-detection, I disabled that for this file, security isn't necessary in isolated tests.

Simon-VdB (approved):
All tests succeed

willemDC-Info (approved):
I looks all good to me and all tests seem to pass so for me it can be merged.

PR #169, opened 2026-03-06
production endpoint

### LINKED ISSUE #14: [DEV] API endpoints: productions
**Location**

`/backend/src/routes/productions/`

____

**Description**

Implement the productions endpoint of the API. (production + custom_production_field database models)

`/api/v1/productions/`

____

**Sources**

### LINKED ISSUE #21: [DEV] productions endpoint: language support
**Location**

`/backend/src/routes/`

____

**Description**

Add support for dutch and english to the productions endpoint of the API. (`...?lang={en/nl}` search param)

____

**Sources**

### LINKED ISSUE #22: [DEV] productions endpoint: fetch productions list
**Location**

`/backend/src/routes/`

____

**Description**

Create a function to fetch a list of productions. This shouldn't include all details of a productions, only the important ones for a list view. Also include support for search queries, filter, sort and paging.

Handle `GET /api/v1/productions?{search, filter and sort params}`.

____

**Sources**

See mock-up for the important fields needed in a list view and the necessary params for search, filtering and sorting.

### LINKED ISSUE #23: [DEV] productions endpoint: fetch production
**Location**

`/backend/src/routes/`

____

**Description**

Create a function to fetch a production including all details (which will be used for the detail page).

Handle `GET /api/v1/productions/{id}`.

____

**Sources**

### LINKED ISSUE #25: [DEV] productions endpoint: create production
**Location**

`/backend/src/routes/`

____

**Description**

Implement a function to create a production.

Handle `POST /api/v1/productions` with a json body of the production.

____

**Sources**

### LINKED ISSUE #26: [DEV] productions endpoint: edit production
**Location**

`/backend/src/routes/`

____

**Description**

Create a function to edit a production. Handle `PUT /api/v1/productions/{id}` and `PATCH /api/v1/productions/{id}` with a json body of the editted fields.

____

**Sources**

### LINKED ISSUE #27: [DEV] productions endpoint: bulk edit productions
**Location**

`/backend/src/routes/`

____

**Description**

Implement a bulk edit of productions.

Not sure what HTTP request method we will use. There are 2 options:
- do a PUT or PATCH for each production that is edited (simple, but slower)
- do a PUT or PATCH to `/api/v1/productions/bulk` with a body that has a set of id's and the fields that need to be edited (more complex, but faster)

Also something to think about is how errors will be handled. There are 2 options for this as well:
- if one fails, fail all
- if one fails, only this fails (seems better at first glance, but it's less user friendly, the user will have to look for that one failed production to retry)

____

**Sources**

### LINKED ISSUE #28: [DEV] productions endpoint: delete production
**Location**

`/backend/src/routes/`

____

**Description**

Implement a function to delete a production.

Handle `DELETE /api/v1/productions/{id}`.

____

**Sources**

lucasodb:
**Issue(s)**

closes #21, #22, #23, #25, #26, #27, #28 

____

**Description**

Implemented the production endpoint.

Language support and tests have not been added yet.

____

**Sources**


lucasodb:
Coverage is at 100% for this branch, but no tests have been written yet for the `feat/endpoint-helpers` branch. Once this is done and merged into this branch the validation will pass.

backend/src/routes/production/handlers/bulk-edit.ts
```diff
@@ -0,0 +1,116 @@
...
+    description: ProductionShape["description"]!.optional(),
+    description_extra: ProductionShape["description_extra"]!.optional(),
+    description_2: ProductionShape["description_2"]!.optional(),
+    video_1: ProductionShape["video_1"]!.optional(),
+    video_2: ProductionShape["video_2"]!.optional(),
+    quote: ProductionShape["quote"]!.optional(),
+    quote_source: ProductionShape["quote_source"]!.optional(),
+    programme: ProductionShape["programme"]!.optional(),
+    info: ProductionShape["info"]!.optional(),
+  }).partial();
```
noahvsb: Use the `EditProductionBodySchema` from `edit.ts`

backend/src/routes/production/handlers/bulk-edit.ts
```diff
@@ -0,0 +1,116 @@
...
+  }
+  if (Object.prototype.hasOwnProperty.call(data, "quote_source")) {
+    addField("quote_source", data.quote_source ?? null);
+  }
+  if (Object.prototype.hasOwnProperty.call(data, "programme")) {
+    addField("programme", data.programme ?? null);
+  }
+  if (Object.prototype.hasOwnProperty.call(data, "info")) {
+    addField("info", data.info ?? null);
+  }
```
noahvsb, line 68: You could save each field in an array and go over them in a loop instead of 14 if's.

backend/src/routes/production/handlers/create.ts
```diff
@@ -0,0 +1,85 @@
...
+  addField("teaser", body.teaser);
+  addField("description", body.description ?? null);
+  addField("description_extra", body.description_extra ?? null);
+  addField("description_2", body.description_2 ?? null);
+  addField("video_1", body.video_1 ?? null);
+  addField("video_2", body.video_2 ?? null);
+  addField("quote", body.quote ?? null);
+  addField("quote_source", body.quote_source ?? null);
+  addField("programme", body.programme ?? null);
+  addField("info", body.info ?? null);
```
noahvsb: Here you can also use a list and loop.

backend/src/routes/production/handlers/edit.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+  if (Object.prototype.hasOwnProperty.call(body, "quote_source")) {
+    addField("quote_source", body.quote_source ?? null);
+  }
+  if (Object.prototype.hasOwnProperty.call(body, "programme")) {
+    addField("programme", body.programme ?? null);
+  }
+  if (Object.prototype.hasOwnProperty.call(body, "info")) {
+    addField("info", body.info ?? null);
+  }
```
noahvsb, line 64: Here also use a list and loop.

backend/src/routes/production/production.ts
```diff
@@ -0,0 +1,33 @@
...
+ * - `PATCH /api/v1/production/bulk` — bulk update multiple productions.
+ * - `DELETE /api/v1/production/:id` — delete a production by ID.
+ *
+ * @param server - The Fastify instance to register routes on.
+ */
+export default function productionRoutes(server: FastifyInstance) {
+  server.get("/api/v1/production", replyHandler(server, fetchProductions));
+  server.get("/api/v1/production/:id", replyHandler(server, fetchProduction));
+  server.post("/api/v1/production", replyHandler(server, createProduction));
+  server.patch("/api/v1/production/:id", replyHandler(server, editProduction));
```
noahvsb: I would also add a PUT where you replace the whole production.

Some users might prefer replacing, because it allows you to edit a few fields field of a fetched production and then send that whole production back, instead of extracting only the edited fields.

backend/src/routes/production/production.ts
noahvsb, line 1: Authorization still has to be added. But I would wait until #168 has been approved and merged. Because we might still change how authorization works.

noahvsb (changes requested):
For your body schema's, you can use `Schema.pick()` for a cleaner approach.

backend/src/routes/production/handlers/bulk-edit.ts
noahvsb, line 1: Fix eslint warnings

backend/src/routes/production/handlers/create.ts
noahvsb, line 1: Fix eslint warnings

backend/src/routes/production/handlers/edit.ts
noahvsb, line 1: Fix eslint warnings

backend/src/routes/production/handlers/replace.ts
noahvsb, line 1: Fix eslint warnings

noahvsb (changes requested):
Looks good to me, besides the lack of authorization (which you can't do anything about yet) and the eslint warnings.

backend/src/routes/helpers.ts
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * Returns metadata for database operations.
+ * Until authorization is implemented, `admin` is hardcoded to `0`.
+ *
+ * @returns An object containing the current admin ID and the current timestamp.
+ */
+export function getMetadata() {
+  return {
+    admin: 0,
```
Simon-VdB: this is a placeholder I assume?

backend/src/routes/production/handlers/bulk-edit.ts
```diff
@@ -0,0 +1,83 @@
...
+  fields.push(`updated_by = $${i++}`, `updated_at = $${i++}`);
+  values.push(admin, current_time, ids);
+
+  await server.pg.query(
+    `UPDATE production SET ${fields.join(", ")} WHERE id = ANY($${i})
+      RETURNING id`,
+    values,
+  );
+
+  const updatedProductions = await Promise.all(ids.map((id) => getProductionById(server, id as string | number)));
```
Simon-VdB: This does N individual select queries. Maybe it would be beter to do one select:
WHERE p.id = ANY($1)
?

backend/src/routes/production/handlers/edit.ts
```diff
@@ -0,0 +1,75 @@
...
+  fields.push(`updated_by = $${i++}`, `updated_at = $${i++}`);
+  values.push(admin, current_time, id);
+
+  await server.pg.query(
+    `UPDATE production SET ${fields.join(", ")} WHERE id = $${i}
+     RETURNING id`,
+    values,
+  );
+
+  return await getProductionById(server, id);
```
Simon-VdB, line 80: If an empty body is given than you will have a no-op update that changes the metadata and returns a 200 OK while nothing happend. This shouldn't be possible.

Simon-VdB (commented):
Tests succeed, just have to wait for helpers and auth.

backend/src/routes/production/handlers/bulk-edit.ts
lucasodb: This is a valid remark, I'll implement a helper for this in the fetch endpoint.

backend/src/routes/production/handlers/edit.ts
lucasodb: Yes, when a patch is done with and empty body the metadata should not be updated. I will make it so an actual change to the fields is required.

backend/src/routes/production/handlers/bulk-edit.ts
```diff
@@ -0,0 +1,81 @@
...
+    if (hasOwn(data, column)) {
+      addField(column, getNullableFieldValue(data, column));
+    }
+  }
+
+  // Always update metadata
+  fields.push(`updated_by = $${i++}`, `updated_at = $${i++}`);
+  values.push(admin, current_time, ids);
+
+  await server.pg.query(
```
Simon-VdB, line 78: maybe use the query builder here? it's optional it's just a helper function, but I just want to address it.

backend/src/routes/production/handlers/bulk-edit.ts
Simon-VdB, line 1: In edit.ts you've done an excellent job of making sure you cannot change the metadata without changing any fields by throwing a "No fields to update" error. I would love to see that in the bulk edit as well.

Simon-VdB (dismissed):

backend/test/routes/production/fetch.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      url: `/api/v1/production/${baseProduction["id"]}`,
+      cookies: { session: sessionCookie },
+    });
+
+    expect(response.statusCode).toBe(200);
+    const parsed = ProductionSchema.parse(response.json());
+    expect(parsed).toEqual(ProductionSchema.parse(baseProduction));
+  });
+
+  test("GET /api/v1/production/:id -> returns 404 for unknown id", async () => {
```
msth5, line 100: Je zou nog kunnen testen op invalid ID (IDs die dus geen nummer zijn), parseParams heeft da normaal gezien, maar ik zou het nog bij alle methodes testen die een ID in de URI heeft

backend/src/routes/production/production.ts
```diff
@@ -0,0 +1,38 @@
...
+ * - `PATCH /api/v1/production/:id` — partially update an existing production.
+ * - `PATCH /api/v1/production/bulk` — bulk update multiple productions.
+ * - `DELETE /api/v1/production/:id` — delete a production by ID.
+ *
+ * @param server - The Fastify instance to register routes on.
+ */
+export default function productionRoutes(server: FastifyInstance) {
+  const protect = { preHandler: [server.authorize] };
+
+  server.get("/api/v1/production", protect, replyHandler(server, fetchProductions));
```
msth5: Correct me if I'm wrong, but I thought the gets (without metadata) were unprotected.
Also, is there no way to get metadata here?

backend/test/routes/production/create.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+afterAll(async () => {
+  await server.close();
+});
+
+beforeEach(() => {
+  vi.restoreAllMocks();
+});
+
+describe("Create on production route", () => {
```
msth5, line 42: It is an idea to also add at test to check unautherised access doesn't happen (so a test without sessioncookie and it should fail with 401) (apply this to all protected methods if you decide to do)

backend/src/routes/production/handlers/bulk-edit.ts
lucasodb: I don't find this very practical right now, so I won't be using this for the moment, but I have it written down to look at in the future.

backend/test/routes/production/fetch.test.ts
lucasodb: This is not really something specific for this endpoint, so I don't think this is really usefull to test here.

backend/test/routes/production/create.test.ts
lucasodb: Again, not perse usefull and rather something general than specific to this endpoint.

msth5: This is however something that can be different from endpoint to endpoint and thus I'd argue this should be tested per endpoint when applicable
You can test the way how you reach this, but one should also test it happens where necessary

backend/src/routes/production/handlers/fetch.ts
noahvsb, line 1: I would also add a fetch that fetches the metadata as well.

noahvsb (changes requested):

backend/test/routes/production/create.test.ts
noahvsb: I have tested the authorization extensively on my auth branch, it's not necessary to test it here as well.

noahvsb: > This is however something that can be different from endpoint to endpoint

It's not different from the authentication endpoint, so it shouldn't be tested imo.

backend/test/routes/production/fetch.test.ts
noahvsb: I would compare this to using an external library and then writing tests to see if the library throws the errors it says it'll throw, noone does that.

noahvsb (changes requested):
You should delete [/backend/src/routes/productions.ts](https://github.com/SELab-2/viernulvier-1/blob/9d963e15f56d3d92df0a55f6ecc93fe780921440/backend/src/routes/productions.ts).

And add your routes to [/backend/src/routes/registerRoutes.ts](https://github.com/SELab-2/viernulvier-1/blob/9d963e15f56d3d92df0a55f6ecc93fe780921440/backend/src/routes/registerRoutes.ts). This way you also don't need to register your routes to the server on each test.

noahvsb (approved):
Changes are good.

PR #170, opened 2026-03-07
fix shared imports when using vitest

noahvsb:
**Issue(s)**

____

**Description**

Imports from the shared folder didn't work when running tests with vitest. This fixes it (you can verify by running the tests on the feat/auth-endpoint branch).

____

**Sources**

### DIFF
--- a/backend/vitest.config.ts
+++ b/backend/vitest.config.ts
@@ -5,6 +5,7 @@ export default defineConfig({
   resolve: {
     alias: {
       "@": path.resolve(__dirname, "./src"),
+      "@viernulvier/shared": path.resolve(__dirname, "../shared/src"),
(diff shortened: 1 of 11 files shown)


Simon-VdB:
I don't get why I should test another branch to see if the changes made on this branch are good?

noahvsb:
> I don't get why I should test another branch to see if the changes made on this branch are good?

Because currently there are no tests using imports from the shared folder on staging.

The only tests using it are on my feat/auth-endpoint branch, but I don't want to merge those with staging yet, it's not ready.

Simon-VdB (approved):
The tests succeed. I get that I had to run the tests on a branch which uses imports from shared.

PR #171, opened 2026-03-08
Fix for types with circular dependencies

Ikerstreamer:
**Issue(s)**

N/A

____

**Description**

I made it so all foreign references now reference a schema that doesnt have backward refs as it was causing inifnite recurssion in the type chekcer.

____

**Sources**

N/A


noahvsb:
It's already in `feat/endpoint-helpers`

PR #180, opened 2026-03-10
Feat/event endpoint

msth5:
**Issue(s)**

Closes #30, Closes #31, Closes #32, Closes #33, Closes #34, Closes #35

____

**Description**

All endpoints for event except language support
New branch so no PR

____

**Sources**

PR #181, opened 2026-03-11
Feat/event price endpoint

msth5:
**Issue(s)**

no issues created for this endpoint

____

**Description**

Endpoints for event price, table only used by event

____

**Sources**

n/a


msth5:
This is a continuation from the Event endpoint feature branch, files under folders `routes/event/` shouldn't be evaluated here.

backend/src/routes/event_price/event_price.ts
noahvsb, line 1: typo in file name

noahvsb (changes requested):
After you fix the indent issue on #180, merge with this branch and fix it here as well.

backend/src/routes/event_price/event_pice.ts
```diff
@@ -0,0 +1,24 @@
...
+    fetchEventPrices, 
+    fetchEventPriceWithMeta, 
+    replaceEventPrice, 
+    editEventPrice,
+ } from "./handlers/index.js";
+
+export default function eventPriceRoutes(server: FastifyInstance) {
+    const protect = { preHandler: [server.authorize] };
+
+    server.get("/event_price/:id", replyHandler(server, fetchEventPrice));
```
trherber: Shouldn't the routes be prefixed with `/api/v1/`?

trherber (commented):
Looks good to me! The only thing I noticed is that you're still using the old helper functions that are now deprecated. `parse` is now `parseSchema` (even though they're just aliases) and `parseFirstRow` is also deprecated, `buildQuery` is now recommended instead but there was some discussion about `buildQuery` so I'm not sure whether it should be changed.

backend/src/routes/event_price/event_pice.ts
msth5: yes indeed

backend/migrations/003.do.create-event-tables.sql
```diff
@@ -40,12 +40,15 @@ CREATE INDEX IF NOT EXISTS idx_event_hall_starts_at ON event(hall, starts_at);
 -- ============================================================
 -- event is both PK and FK: exactly one price record per event (1:1 extension).
 CREATE TABLE IF NOT EXISTS event_price (
-  event           INT           PRIMARY KEY REFERENCES event(id) ON DELETE CASCADE,
+  id              INT           PRIMARY KEY,
```
Simon-VdB: add "SERIAL" here for key generation

Simon-VdB (dismissed):

noahvsb (changes requested):
Here as well, it seems that I was tripping, indents are good.

noahvsb (approved):
Forgot to click approve on my previous review.

PR #184, opened 2026-03-12
fixed backend error when building docker

Simon-VdB:
**Issue(s)**

N/A

____

**Description**

Backend had an error when building. This PR fixes it.

____

**Sources**

N/A


.claude/worktrees/elastic-williamson
noahvsb, line 1: oops

noahvsb (commented):
The error can't originate from admin, because it worked after admin was pushed.

noahvsb:
> The error can't originate from admin, because it worked after admin was pushed.

It did fix it weirdly enough, I just tested the wrong way. The error only throws in the dev environment, so I thought it was already fixed by Lex's PR, but I tested it in the build, not the dev environment.

noahvsb (approved):
I tested both the non-dev and dev environments and both work.

PR #185, opened 2026-03-12
fixed inconsistencies in shared module

noahvsb:
**Issue(s)**

____

**Description**

I found some inconsistencies in the shared module that could result in problems later down the line.

____

**Sources**


noahvsb:
@Simon-VdB If you add these inconsistency fixes to your backend build fix, then the backend build can be merged.

PR #186, opened 2026-03-12
fix TSC linking issues in build

Acher0s:
**Issue(s)**

N/A

____

**Description**

added tool that will fix linking issues in app build

____

**Sources**

https://alberthernandez.dev/blog/typescript-path-aliases-in-node-js-the-complete-guide


noahvsb (approved):
Looks good to me, glad it works now.

PR #202, opened 2026-03-13
Feat/new runners

Acher0s:
**Issue(s)**

N/A

____

**Description**

New action runnners were added but they are x64, so workflow file needs to allow those to run them too.

____

**Sources**

N/A


.github/workflows/pr-staging.yml
```diff
@@ -24,8 +24,8 @@ jobs:
         working-directory: ${{ matrix.project }}
 
     steps:
-      - name: Clear Docker build cache
-        run: docker builder prune -af
+      # - name: Clear Docker build cache
+      #   run: docker builder prune -af
```
noahvsb, line 28: Why is this commented out?

Acher0s: clearing cache was a temporary solution when a cache corruption happend, it shouldn't be there under normal circumstances.

noahvsb (approved):
Thanks for clearing it up

PR #205, opened 2026-03-15
Feat/router setup

### LINKED ISSUE #106: [DEV] Public archive: Router setup
**Location**

`/frontend/src/router/

____

**Description**

Set up the application router with all required routes for the public archive website. This includes routes for: the list view (home/agenda), production detail page, series (tag) detail page, and the print uploads page. The router should support language prefixes (e.g. /nl/productions, /en/productions).

____

**Sources**

- Requirements viernulvier

trherber:
**Issue(s)**

Closes #106  — [DEV] Public archive: Router setup

____

**Description**

Sets up the application router for the public archive website with the following:
- Routes for home, productions list view, production detail and admin cms, all with language prefix support (/nl, /fr, /en)
- Redirects for routes without a language prefix (e.g. /productions → /nl/productions) 
- Automatic language detection based on localStorage (saved preference), browser language, and a default fallback
- vue-i18n integration — the active locale (currently active language) is synced with the URL on every navigation
- Admin route guard for /admin/cms
- 404 fallback for unknown routes
____

**Sources**


frontend/src/router/index.ts
```diff
@@ -0,0 +1,56 @@
...
+    if (matchedRoute.name !== RouteNames.NOT_FOUND) {
+      return next(pathWithLang);
+    }
+
+    // else, route truly doesn't exist, proceed to 404
+    return next();
+  }
+
+  // set locale based on route param
+  i18n.global.locale.value = to.params.lang as SupportedLang;
```
Simon-VdB:
```suggestion
  if (to.params.lang) {
    i18n.global.locale.value = to.params.lang as SupportedLang;
  }
```

frontend/src/i18n/index.ts
```diff
@@ -0,0 +1,45 @@
...
+ * so it can be restored on the next visit.
+ */
+export function saveLanguagePreference(lang: SupportedLang): void {
+  localStorage.setItem(STORAGE_KEY, lang);
+}
+
+/** The global vue-i18n instance. Import `i18n` to access or change the active locale. */
+export const i18n = createI18n({
+  legacy: false,
+  locale: DEFAULT_LANG,
```
Simon-VdB, line 42:
```suggestion
  locale: detectLanguage(),
```

frontend/src/views/HomeView.vue
```diff
@@ -0,0 +1,131 @@
+<template>
+  <div class="app" :class="{ dark: isDark }">
+    <nav class="navbar">
+      <div class="nav-left">
+        <a href="#" class="nav-link">{{ t("nav.home") }}</a>
```
Simon-VdB: these #'s should later be replaced by the routers, but this is more of a detail for design later on.

frontend/test/composables/useLocale.test.ts
```diff
@@ -0,0 +1,158 @@
...
+        expect(i18n.global.locale.value).toBe(lang);
+      },
+    );
+  });
+
+  // ── Router navigation ──────────────────────────────────────────────────────
+  // Verifies that setLocale() replaces the language prefix in the URL
+  // while preserving the rest of the path (e.g. /nl/productions → /fr/productions).
+
+  describe("router navigatie", () => {
```
Simon-VdB:
```suggestion
  describe("router navigation", () => {
```

frontend/test/composables/useLocale.test.ts
```diff
@@ -0,0 +1,158 @@
...
+        await flushPromises();
+        expect(router.currentRoute.value.path).toBe(`/${lang}`);
+      },
+    );
+  });
+
+  // ── All together ───────────────────────────────────────────────────────────
+  // Verifies that all three side effects fire correctly in a single setLocale() call.
+
+  describe("alles synchroon", () => {
```
Simon-VdB:
```suggestion
  describe("everything synchronous", () => {
```

Simon-VdB (changes requested):
Few minor changes, looks good overall. One question: I assume frontend/src/api and frontend/src/util are folders where helpers would go in the future? On the helpers branch I've named them slightly different, and that's no problem, but I'd say for separation of concern's sake that we don't force any naming as placeholders, if that makes any sense?

frontend/src/router/index.ts
trherber: Defensive programming is indeed cleaner, but since all valid routes always have a language prefix the falsy branch would never be hit. So I would have no test for it either, which drops branch coverage to 90%, is that okay?

frontend/src/i18n/index.ts
trherber: The guard already syncs the locale to the detected language immediately, so changing this here has little effect. It also ties the tests to the jsdom environment language. Not sure it's worth the change.

frontend/src/router/index.ts
Simon-VdB: Hmm, maybe coverage is more important then...

Simon-VdB (dismissed):

Simon-VdB (approved):

PR #212, opened 2026-03-20
Scripts for importing legacy CSV data files 

### LINKED ISSUE #52: [DEV] data-import (CSV)
**Location**

`/backend/src/import/ (?)`

____

**Description**

Implement functions to import a CSV of data into the database.

____

**Sources**

### LINKED ISSUE #53: [DEV] data-import (CSV): parse and validate CSV
**Location**

`/backend/src/import/ (?)`

____

**Description**

Implement a function to parse and validate a CSV file.

____

**Sources**

### LINKED ISSUE #54: [DEV] data-import (CSV): map parsed CSV to database model
**Location**

`/backend/src/import/ (?)`

____

**Description**

Map the parsed (and validated) CSV to our database model.

____

**Sources**

### LINKED ISSUE #55: [DEV] data-import (CSV): import parsed and mapped CSV to database
**Location**

`/backend/src/import/ (?)`

____

**Description**

Implement a function that imports all of this data into our database.

____

**Sources**

lucasodb:
**Issue(s)**

Closes #53, #54, #55

____

**Description**

- Removed unnecessary box_office_id and vendor_id fields from schema's.
  (this change accounts for 40 of the changed files, so don't worry, you don't have to look at that many files)
- Fixed the use of incorrect database table names in event handlers.
- Added scripts for importing legacy production and event data files, expecting the format of the two provided files (https://github.com/SELab-2/viernulvier-opgave/blob/main/Productions%20-%20output.csv and https://github.com/SELab-2/viernulvier-opgave/blob/main/Events%20-%20voorstellingen.csv).

The data of the two provided files were inserted into the database using the scripts, and this was the result:

**Productions:** 

  Rows read: 6129
  Imported productions: 5672
  Created genre tags: 36
  Linked production_tag rows: 5849
  Skipped (already imported): 0
  Skipped (missing legacy ID): 244
  Skipped (duplicate legacy ID in file): 0
  Skipped (missing title): 213
  Failed rows: 0

5672 productions inserted (244 missing legacy ID, 213 missing title skipped).
6129 logical CSV rows processed.

**Events:**

  Rows read: 10598
  Imported events: 10220
  Created halls: 68
  Skipped (missing starttime): 1
  Skipped (missing hall): 94
  Skipped (missing production legacy id): 0
  Skipped (unknown production mapping): 227
  Skipped (duplicate in file): 56
  Skipped (already imported): 0
  Failed rows: 0

10220 events inserted.
10598 rows read; skips: 227 unknown production, 94 missing hall, 56 duplicate in file, 1 missing starttime, etc.
____

**Sources**


data/imports/README.md
noahvsb, line 1: Link to this README in the root's README.md

backend/scripts/import-productions-legacy.ts
noahvsb, line 1: A lot of code is shared between the productions and the events import, please put these in a different file.

data/imports/README.md
```diff
@@ -0,0 +1,70 @@
...
+  - `cd backend && DATABASE_URL="postgres://postgres@localhost:5432/postgres" pnpm run import:productions -- --file "../data/imports/Productions - output.csv"`
+- Event import with default path:
+  - `cd backend && DATABASE_URL="postgres://postgres@localhost:5432/postgres" pnpm run import:events`
+- Event import with custom path:
+  - `cd backend && DATABASE_URL="postgres://postgres@localhost:5432/postgres" pnpm run import:events -- "../data/imports/Events - voorstellingen.csv"`
+
+Optional flags:
+
+- `--dry-run`
+- `--limit <n>`
```
noahvsb: Can you explain these further? It might also be nice to have the help output here in this file, since they are explained there.

noahvsb (changes requested):
The scripts folder isn't tested, because they use functions from the `src` folder that are tested. So move most of the functionality to `src/db` and write tests.

For the parseArgs and printHelp, a cleaner approach would be using the [yargs package](https://www.npmjs.com/package/yargs).

Also can you look into the `DATABASE_URL`? It shouldn't be needed in the command, it's in the `.env`-file. The migrate script also uses the `DATABASE_URL`, but you don't need to add it to the pnpm command.

Running the commands locally works 👍

lucasodb:
> The scripts folder isn't tested, because they use functions from the `src` folder that are tested. So move most of the functionality to `src/db` and write tests.
> 
> For the parseArgs and printHelp, a cleaner approach would be using the [yargs package](https://www.npmjs.com/package/yargs).
> 
> Also can you look into the `DATABASE_URL`? It shouldn't be needed in the command, it's in the `.env`-file. The migrate script also uses the `DATABASE_URL`, but you don't need to add it to the pnpm command.
> 
> Running the commands locally works 👍

- I agree on deduplicating the shared helpers into a module. Makes sense that these also would be tested, but I wouldn't put them under src/db, maybe something like src/legacy-import would be cleaner?

- I don't really think we need to use a package for displaying the flag info,  these parsers are still quite small, but if it's preferred to do this everywhere for consistency then I'm happy to look into it.

- The `DATABASE_URL` was used to override the config in .env mostly for debugging, but as I've got everything to work the way it should with the default setup this indeed does not need to explicitly be used in the command.

noahvsb:
> * I agree on deduplicating the shared helpers into a module. Makes sense that these also would be tested, but I wouldn't put them under src/db, maybe something like src/legacy-import would be cleaner?

That's indeed better.

> * I don't really think we need to use a package for displaying the flag info,  these parsers are still quite small, but if it's preferred to do this everywhere for consistency then I'm happy to look into it.

The package makes it cleaner, does extra checks and it means you don't need to write tests for your `parseArguments` and `printHelp` functions.

> * The `DATABASE_URL` was used to override the config in .env mostly for debugging, but as I've got everything to work the way it should with the default setup this indeed does not need to explicitly be used in the command.

I tried not using it in the command and it threw an error that `DATABASE_URL` was not defined.

lucasodb:
> > * I agree on deduplicating the shared helpers into a module. Makes sense that these also would be tested, but I wouldn't put them under src/db, maybe something like src/legacy-import would be cleaner?
> 
> That's indeed better.
> 
> > * I don't really think we need to use a package for displaying the flag info,  these parsers are still quite small, but if it's preferred to do this everywhere for consistency then I'm happy to look into it.
> 
> The package makes it cleaner, does extra checks and it means you don't need to write tests for your `parseArguments` and `printHelp` functions.
> 
> > * The `DATABASE_URL` was used to override the config in .env mostly for debugging, but as I've got everything to work the way it should with the default setup this indeed does not need to explicitly be used in the command.
> 
> I tried not using it in the command and it threw an error that `DATABASE_URL` was not defined.

Yes, working on this but something else came up and I tried to fix that first and forgot to actually make this work, will soon push this.

I'll also look into Yargs.

README.md
```diff
@@ -6,18 +6,22 @@ VIERNULVIER has been a cornerstone of Ghent's cultural scene for over a century,
...
 ## 🚀 Getting Started
 
 Refer to the **[Contributing Guide](DOCS/CONTRIBUTING.md)** for full setup instructions, including how to run the application locally and contribute to the project.
 
 ## 🔗 Links
 
-- **Live archive:** https://viernulvier-archive.be
-- **Issues & feedback:** use the GitHub issue tracker
\ No newline at end of file
+- **Live archive:** [https://viernulvier-archive.be](https://viernulvier-archive.be)
```
noahvsb: what's this?

noahvsb (changes requested):
Mostly looks good to me. But I think the script files are still too big to leave untested. So move almost all logic to `/src/legacy-import/{something with event}.ts` and idem for production. Rename the `main` function to  `import{Events|Productions}Legacy`, which you export and use in the script. It's the same approach as Iker used in his migrate script.

lucasodb:
> Mostly looks good to me. But I think the script files are still too big to leave untested. So move almost all logic to `/src/legacy-import/{something with event}.ts` and idem for production. Rename the `main` function to `import{Events|Productions}Legacy`, which you export and use in the script. It's the same approach as Iker used in his migrate script.

This is fair. However, because of the nature of the code (for example, "every 500 rows, log progress" logic, which means you would have to process 500 data rows in one test) achieving the set coverage percentage would be very expensive, so I exluded this code from the global 97.5% gate. I have still aimed for about 90% coverage though.

README.md
lucasodb: What do you mean? This is the domain that Lex setup to run our app on. I thought I would put the link there.

noahvsb: The part between square brackets is the same as between the round brackets.

lucasodb: Uhh oops did not mean for this.

backend/vitest.config.ts
```diff
@@ -16,7 +16,11 @@ export default defineConfig({
     coverage: {
       provider: "v8",
       include: ["src/**/*.{ts,tsx}"],
-      exclude: ["src/index.ts"],
+      exclude: [
+        "src/index.ts",
+        "src/legacy-import/import-productions-legacy.ts",
+        "src/legacy-import/import-events-legacy.ts",
```
noahvsb: why are they excluded?

noahvsb:
> so I exluded this code from the global 97.5% gate. I have still aimed for about 90% coverage though.

Oh okay, I see. But you don't need to exclude it, you can just add a different coverage threshold for these files.

backend/vitest.config.ts
noahvsb, line 41:
```suggestion
    silent: true,
  },
```

Since the console.logs are getting longer and longer, I think it's best to just silence them when running the tests.

noahvsb (approved):
I tested again locally and it still works.

PR #219, opened 2026-03-21
add support for authorization header usage

### LINKED ISSUE #177: [DEV] allow authorization using authorization header instead of only with cookies
**Location**

`/backend/src/plugins/authorize.ts`

____

**Description**

Currently the only option to authorize is using cookies. For completeness, it would be nice to also have support for the authorization header.

____

**Sources**

You should look up the HTTP spec and information on @fastify/jwt and @fastify/cookie.

### LINKED ISSUE #203: [BUG] logout only deleted the session cookie locally
**Description**

The problem with only deleting locally is that if a user made a back-up and inserts it back in the browser, the user can still authorize with that same cookie. So it should be deleted server-side as well.

____

**Environment**

curl

**Sources**

noahvsb:
**Issue(s)**

closes #177 
due to the way I changed the logout I also resolved #203, it now revokes the token when doing logout, instead of only clearing the session cookie locally.

____

**Description**

Implemented the authorization header.

You can test it with curl (after setting up docker and running migrate + create-admin scripts):

Login:
```sh
curl -X POST http://localhost:5173/api/v1/auth/login -H "Content-Type: application/json" -d '{"username":"admin","password":"password"}'
```
That will output a token, store it in a shell variable:
```sh
TOKEN=<copy paste>
```
Fetch mock admin to see if it works:
```sh
curl -H "Authorization: Bearer $TOKEN" http://localhost:5173/api/v1/auth/1
```
Logout:
```sh
curl -X POST -H "Authorization: Bearer $TOKEN" http://localhost:5173/api/v1/auth/logout
```
Fetch mock admin should say "Token has been revoked":
```sh
curl -H "Authorization: Bearer $TOKEN" http://localhost:5173/api/v1/auth/1
```

____

**Sources**

### DIFF
--- a/backend/src/plugins/jwt.ts
+++ b/backend/src/plugins/jwt.ts
@@ -2,11 +2,18 @@ import fp from "fastify-plugin";
 import type { FastifyInstance } from "fastify";
 import fastifyJwt from "@fastify/jwt";
 import fastifyCookie from "@fastify/cookie";
+import { randomUUID } from "crypto";
+
+const denylist = new Set<string>();
 
 /**
  * Registers JWT and cookie plugins on the Fastify instance.
  * The JWT secret is read from the `JWT_SECRET` environment variable.
  * Sessions are stored in an `httpOnly` cookie named `session`.
+ *
+ * Also decorates the server with:
+ * - `tokenDenylist` — a Set of revoked JWT IDs (`jti` claims) that are rejected on each request.
+ * - `generateJti` — a function that generates a unique JWT ID for each token.
  *
  * @param server - The Fastify instance to register the plugins on.
  */
@@ -26,4 +33,14 @@ export default fp(async function jwtPlugin(server: FastifyInstance) {
       signed: false,
     },
   });
+
+  server.decorate("tokenDenylist", denylist);
+  server.decorate("generateJti", () => randomUUID());
 });
+
+declare module "fastify" {
+  interface FastifyInstance {
+    tokenDenylist: Set<string>;
+    generateJti: () => string;
+  }
+}
--- a/backend/src/routes/auth/handlers/login.ts
+++ b/backend/src/routes/auth/handlers/login.ts
@@ -25,12 +25,14 @@ const fetchAdminCredentials = (server: FastifyInstance) =>
 const DUMMY_HASH = "$2b$12$invalidhashvaluethatwillnevermatchangything";
 
 /**
- * Authenticates an admin by username and password and sets a session cookie.
+ * Authenticates an admin by username and password, sets a session cookie, and returns a signed JWT.
+ * The token contains the admin's `id`, `username`, and a unique `jti` claim used for revocation on logout.
+ * Clients that cannot use cookies should store the returned token and pass it via the `Authorization: Bearer <token>` header.
  *
  * @param server - The Fastify instance, used for database access and logging.
  * @param request - The Fastify request, expected to contain `username` and `password` in its body.
  * @param reply - The Fastify reply, used to set the session cookie.
- * @returns An object indicating success.
+ * @returns The signed JWT token.
  * @throws `HttpError` With status 401 if the username is not found or the password is incorrect.
  */
 export async function login(server: FastifyInstance, request: FastifyRequest, reply: FastifyReply) {
@@ -38,13 +40,12 @@ export async function login(server: FastifyInstance, request: FastifyRequest, re
 
   const rows = await fetchAdminCredentials(server)(username);
 
-  // always compare hash to prevent a timing attack
   const valid = await comparePassword(password, rows[0]?.password ?? DUMMY_HASH);
 
   if (!valid || rows.length == 0) throw new HttpError(401, "Invalid credentials");
 
   const token = server.jwt.sign(
-    { id: rows[0]!.id, username },
+    { id: rows[0]!.id, username, jti: server.generateJti() },
     { expiresIn: "24h" },
   );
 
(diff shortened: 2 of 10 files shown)


noahvsb:
I keep mixing up authentication and authorization

backend/src/routes/auth/handlers/logout.ts
```diff
@@ -1,14 +1,32 @@
...
- * @param _request - Unused.
+ * @param server - The Fastify instance, used to access the token denylist.
+ * @param request - The Fastify request, used to extract and verify the session token.
  * @param reply - The Fastify reply, used to clear the session cookie.
  * @returns An object indicating success.
  */
-export async function logout(_server: FastifyInstance, _request: FastifyRequest, reply: FastifyReply) {
+export async function logout(server: FastifyInstance, request: FastifyRequest, reply: FastifyReply) {
+  try {
+    const payload = await request.jwtVerify() as { jti?: string; exp?: number };
```
Simon-VdB, line 15: This is a second jwtVerify call. Logout is now a protected route which means it already does jwtVerify in the authorizePlugin. I think you can replace this with:

```suggestion
    const payload = request.user as { jti?: string; exp?: number };
```

backend/src/routes/auth/handlers/logout.ts
```diff
@@ -1,14 +1,32 @@
...
- * @param _server - Unused.
- * @param _request - Unused.
+ * @param server - The Fastify instance, used to access the token denylist.
+ * @param request - The Fastify request, used to extract and verify the session token.
  * @param reply - The Fastify reply, used to clear the session cookie.
  * @returns An object indicating success.
  */
-export async function logout(_server: FastifyInstance, _request: FastifyRequest, reply: FastifyReply) {
+export async function logout(server: FastifyInstance, request: FastifyRequest, reply: FastifyReply) {
+  try {
```
Simon-VdB, line 14: This try-catch is redundant, because...

backend/src/routes/auth/handlers/login.ts
```diff
@@ -25,26 +25,27 @@ const fetchAdminCredentials = (server: FastifyInstance) =>
...
- * @returns An object indicating success.
+ * @returns The signed JWT token.
  * @throws `HttpError` With status 401 if the username is not found or the password is incorrect.
  */
 export async function login(server: FastifyInstance, request: FastifyRequest, reply: FastifyReply) {
   const { username, password } = parseSchema(server, LoginBodySchema, request.body);
 
   const rows = await fetchAdminCredentials(server)(username);
 
-  // always compare hash to prevent a timing attack
```
Simon-VdB, line 41: why delete this? This was meaningful information...

backend/src/routes/auth/handlers/login.ts
```diff
@@ -25,26 +25,27 @@ const fetchAdminCredentials = (server: FastifyInstance) =>
...
   const valid = await comparePassword(password, rows[0]?.password ?? DUMMY_HASH);
 
   if (!valid || rows.length == 0) throw new HttpError(401, "Invalid credentials");
```
Simon-VdB, line 45: I know this is not you're line of code, but isn't === better for the consistency?

backend/src/plugins/jwt.ts
```diff
@@ -2,12 +2,19 @@ import fp from "fastify-plugin";
 import type { FastifyInstance } from "fastify";
 import fastifyJwt from "@fastify/jwt";
 import fastifyCookie from "@fastify/cookie";
+import { randomUUID } from "crypto";
+
+const denylist = new Set<string>();
```
Simon-VdB, line 7: this lives in process memory... On server restart all revoked tokens are forgotten...

backend/test/routes/auth/logout.test.ts
```diff
@@ -31,10 +40,44 @@ describe("Logout on auth route", () => {
     );
   });
 
-  test("POST /api/v1/auth/logout — succeeds without session cookie (just does nothing)", async () => {
+  test("POST /api/v1/auth/logout — returns 401 when token is revoked", async () => {
```
Simon-VdB, line 43: Tests should be independent — give each test its own freshly signed token.

Simon-VdB (commented):
solid implementation, keep it up.

backend/src/routes/auth/handlers/logout.ts
noahvsb: Correct, now the try catch can also go. I implemented this before realizing it's best to just protect this route.

backend/src/routes/auth/handlers/login.ts
noahvsb: AI :saluting_face: 

I missed it when reviewing the changes.

backend/src/routes/auth/handlers/logout.ts
noahvsb: I should've read the next comment first haha.

backend/src/plugins/jwt.ts
noahvsb: If the server restarts, we should use a new `JWT_SECRET` for security so all tokens are invalid. Then this isn't an issue.

backend/test/routes/auth/logout.test.ts
noahvsb: I agree, but literally all routes tests use the same session cookie per file over every test. I'll change it in this test file because here it's borderline unacceptable, but I would leave the other tests as is.

backend/src/routes/auth/handlers/login.ts
Simon-VdB: No worries brotha

noahvsb:
I accidentally pushed the requested changes to #220, so let's just merge #220 with staging. It will also be less work for the runners.

PR #244, opened 2026-03-28
Fix/tags remove deprecated

lucasodb:
**Issue(s)**

Closes #232 

____

**Description**

Replaced old deprecated helper functions in tag and tag_type handlers by new ones.

____

**Sources**

PR #246, opened 2026-03-28
rename FK columns

### LINKED ISSUE #245: [DEV] Align FK column names with referenced tables
**Location**

Backend DB layer (migrations + SQL), shared types, and docs/import README

____

**Description**

Rename domain FK columns: tag_type, production_tag.production/tag, image, crop, blogpost, production_custom_field

____

**Sources**

lucasodb:
**Issue(s)**

Closes #245 

____

**Description**

Adds migration 009 to rename foreign-key columns so they match the referenced table names to be more consistent (same idea as event.production / event.hall) , and updates backend SQL, shared Zod shapes, legacy import paths, and docs accordingly.

If #244 is merged before this pr, change the base branch into staging.

____

**Sources**

### DIFF
--- a/DOCS/DATABASE.md
+++ b/DOCS/DATABASE.md
@@ -134,18 +134,18 @@ Table event_price {
 
 Table production_tag {
   ~metadata
-  tag_id int [ref: > tag.id]
-  production_id int [ref: > production.id]
+  tag int [ref: > tag.id]
+  production int [ref: > production.id]
 }
 
 Table tag {
   ~metadata
   "id" int [PK]
   "name" varchar(32) [not null]
-  "type" int [ref: > tag_type.id, not null]
+  "tag_type" int [ref: > tag_type.id, not null]
 
   indexes {
-    type
+    tag_type
   }
 }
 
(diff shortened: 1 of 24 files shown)


noahvsb (commented):
It all looks good.

But I get this error for each of the tags after fetching them on http://localhost:5173/api/v1/tags:

```
{
    "expected": "number",
    "code": "invalid_type",
    "path": [
      0,
      "type"
    ],
    "message": "Invalid input: expected number, received undefined"
}
```

Will this be fixed in #247?

lucasodb:
> It all looks good.
> 
> But I get this error for each of the tags after fetching them on http://localhost:5173/api/v1/tags:
> 
> ```
> {
>     "expected": "number",
>     "code": "invalid_type",
>     "path": [
>       0,
>       "type"
>     ],
>     "message": "Invalid input: expected number, received undefined"
> }
> ```
> 
> Will this be fixed in #247?

Yes. This error is because rows have `tag_type`, not `type`, and list rows don’t include productions until we merge from `production_tag`. That’s fixed in #247.

noahvsb (approved):
#247 fixed the issue, so this can be merged now

PR #247, opened 2026-03-28
refactor tag fetch handler to also fetch production ids

### LINKED ISSUE #201: [DEV] tags endpoint: include list of productions that have this tag on fetch
**Location**

`/backend/src/routes/tag/handlers/fetch.js`

____

**Description**

On fetch, the tag json should include a list of production id's that have this tag.

____

**Sources**

lucasodb:
**Issue(s)**

Closes #201 

____

The old code treated query rows like full Tag objects, which doesn’t match the DB and breaks after the rename migration.
We now parse rows with a small DB schema, load production ids from production_tag, and merge into Tag for the response. Lists leave productions empty unless ?includeProductions=true. single-tag routes still return full productions.

If #246  is merged before this pr, change the base branch into staging.

____

**Sources**

### DIFF
--- a/backend/src/routes/tag/handlers/create.ts
+++ b/backend/src/routes/tag/handlers/create.ts
@@ -9,7 +9,7 @@ const CreateTagBodySchema = TagSchema.pick({
   name: true,
   public: true,
 }).extend({
-  type: z.int().nonnegative(),
+  tag_type: z.int().nonnegative(),
 });
 
 const insertTag = (server: FastifyInstance) =>
@@ -31,7 +31,7 @@ export async function createTag(
 
   const rows = await insertTag(server)(
     body.name,
-    body.type,
+    body.tag_type,
     body.public,
     admin,
     current_time,
--- a/backend/src/routes/tag/handlers/create.ts
+++ b/backend/src/routes/tag/handlers/create.ts
@@ -1,16 +1,10 @@
 import type { FastifyInstance, FastifyRequest } from "fastify";
 import type { Tag } from "@viernulvier/shared/index.js";
 import { TagSchema } from "@viernulvier/shared/index.js";
-import { languageMap } from "@viernulvier/shared/types/helpers.js";
 import { getMetadata, parseSchema, buildQuery } from "@/routes/helpers.js";
-import { z } from "zod";
+import z from "zod";
 
-const CreateTagBodySchema = TagSchema.pick({
-  name: true,
-  public: true,
-}).extend({
-  tag_type: z.int().nonnegative(),
-});
+const CreateTagBodySchema = TagSchema.omit({ id: true, productions: true });
 
 const insertTag = (server: FastifyInstance) =>
   buildQuery(
@@ -18,7 +12,13 @@ const insertTag = (server: FastifyInstance) =>
     `INSERT INTO tag (name, tag_type, public, created_by, updated_by, created_at, updated_at)
      VALUES ($1, $2, $3, $4, $4, $5, $5)
      RETURNING id, name, tag_type, public`,
-    z.tuple([languageMap, z.int(), z.boolean(), z.int(), z.date()]),
+    z.tuple([
+      CreateTagBodySchema.shape.name,
+      CreateTagBodySchema.shape.tag_type,
+      CreateTagBodySchema.shape.public,
+      z.int(),
+      z.date(),
+    ]),
     TagSchema,
   );
 
--- a/backend/src/routes/tag/handlers/fetch.ts
+++ b/backend/src/routes/tag/handlers/fetch.ts
@@ -9,20 +9,9 @@ SELECT id, name, tag_type, public
 FROM tag
 `;
 
-/** Row shape returned by tag SELECTs before `productions` is merged from `production_tag`. */
-const TagDbRowSchema = z.object({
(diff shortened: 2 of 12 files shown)


backend/src/routes/tag/handlers/fetch.ts
```diff
@@ -1,32 +1,112 @@
...
+  created_at: z.coerce.date(),
+  updated_at: z.coerce.date(),
+  created_by: z.int().nonnegative(),
+  updated_by: z.int().nonnegative(),
+});
+
 const TagsListQuerySchema = z.object({
   production: stringToInt.optional(),
+  /** When `includeProductions=true`, each tag includes full `productions` id list (default: empty arrays). */
+  includeProductions: z.literal("true").optional(),
```
noahvsb, line 19: I like that you made this optional, it will make it more efficient. But I don't think an empty array is the way to go, it can be confusing. So I would just leave the productions field out if it's false.

backend/src/routes/tag/handlers/fetch.ts
```diff
@@ -1,32 +1,112 @@
...
+import { parseParams, buildQuery, parseSchema, ParseContext } from "@/routes/helpers.js";
 import { z } from "zod";
 
 const TagSelect = `
 SELECT id, name, tag_type, public
 FROM tag
 `;
 
+/** Row shape returned by tag SELECTs before `productions` is merged from `production_tag`. */
+const TagDbRowSchema = z.object({
```
noahvsb: I don't think you need a new zod object, I think you can just do something like `TagSchema.pick({ id, name, type, public })` or `TagSchema.omit({ productions })`. It might give a problem because it's `tag_type` in the database and `type` in the `TagSchema`, so either rename the type in the shared folder or add a new migration file.

backend/src/routes/tag/handlers/fetch.ts
```diff
@@ -1,32 +1,112 @@
...
+  id: TagSchema.shape.id,
+  name: TagSchema.shape.name,
+  tag_type: z.int().nonnegative(),
+  public: TagSchema.shape.public,
+});
+
+const TagDbRowWithMetaSchema = TagDbRowSchema.extend({
```
noahvsb: Same here but with `TagSchema.withMeta().pick/omit`.

backend/src/routes/tag/handlers/fetch.ts
```diff
@@ -1,32 +1,112 @@
...
+  /** When `includeProductions=true`, each tag includes full `productions` id list (default: empty arrays). */
+  includeProductions: z.literal("true").optional(),
+});
+
+const ProductionTagLinkSchema = z.object({
+  tag: z.int(),
+  production: z.int(),
 });
 
+async function fetchProductionIdsByTagIds(
```
noahvsb, line 27: add TSDoc

backend/src/routes/tag/handlers/fetch.ts
```diff
@@ -1,32 +1,112 @@
...
+  const map = new Map<number, number[]>();
+  for (const { tag, production } of links) {
+    const list = map.get(tag) ?? [];
+    list.push(production);
+    map.set(tag, list);
+  }
+  return map;
+}
+
+function tagsFromDbRows(
```
noahvsb, line 55: add TSDoc

backend/src/routes/tag/handlers/fetch.ts
```diff
@@ -1,32 +1,112 @@
...
+  return rows.map((row) => ({
+    id: row.id,
+    name: row.name,
+    type: row.tag_type,
+    public: row.public,
+    productions: byTag.get(row.id) ?? [],
+  }));
+}
+
+function tagsWithMetaFromDbRows(
```
noahvsb, line 71: add TSDoc

noahvsb (changes requested):
Tests succeed and it works locally.

If you want, you can ignore the TSDoc requested changes because of #249.

backend/src/routes/tag/handlers/fetch.ts
lucasodb: True, it is more clear if this field is just ommited.

backend/src/routes/tag/handlers/create.ts
```diff
@@ -9,7 +9,7 @@ const CreateTagBodySchema = TagSchema.pick({
   name: true,
   public: true,
 }).extend({
-  type: z.int().nonnegative(),
+  tag_type: z.int().nonnegative(),
```
noahvsb: Here you can just pick the tag type now. No need for an extend.

noahvsb (changes requested):
Very small requested change.

backend/src/routes/tag/handlers/create.ts
lucasodb: I did try this, but this gave an error, because picking tag_type from TagSchema leaves its type as unknown because ForeignKey is a custom ZodType subclass. You have to restate tag_type as z.int().nonnegative() to restore a number type. If you want I can look if there is a better solution than doing it this way.

noahvsb: The event endpoint has a create schema that uses foreign keys, try looking into that.

noahvsb (approved):
Nice job

PR #223, opened 2026-03-24
Feat/landing page

Simon-VdB:
**Issue(s)**

(https://github.com/SELab-2/viernulvier-1/issues/107)
(https://github.com/SELab-2/viernulvier-1/issues/108)

____

**Description**

This PR builds out the complete frontend foundation for the VierNulVier archive — from a typed API helper layer to the first fully styled public-facing page.

### Frontend API & utility layer (`src/services/`, `src/utils/`)
- `apiFetch` wrapper with typed `ApiError` (status codes, field errors, convenience getters)
- Resource services for all endpoints: `auth`, `productions`, `events`, `halls`, `tags`, `eventPrices`
- Pure utility modules: `date` (formatting), `i18n` (language map helpers), `form` (Zod validation), `misc` (`debounce`, `groupBy`, `sortBy`, …)

### Landing page (`src/components/`, `src/views/HomeView.vue`)
- **AppNavbar** — sticky top bar with logo, nav links, language switcher, dark-mode toggle and admin button; all wired to `vue-router` and `vue-i18n`
- **HeroSection** — two-column hero with headline, subtitle and CTA
- **StatsSection** — four archive statistics
- **BentoGrid / BentoCard** — catalog-highlights grid with themed cards per content type
- **NewsletterSection** — email subscribe form
- **AppFooter** — copyright, privacy/terms/contact links

### i18n (`src/i18n/locales/`)
- Full `nl` / `en` / `fr` translations for every landing-page string, the date utilities (`date.doorsOpen`) and navigation labels

### Tailwind CSS v4
- Added `tailwindcss` + `@tailwindcss/vite` (v4)
- `@import "tailwindcss"` in `style.css` with custom base layer for `html`, `body` and `#app`
- Google Fonts (Inter) + Material Symbols Outlined added to `index.html`

### Infrastructure fixes
- `tsconfig.app.json` — added `@viernulvier/shared` path alias so shared Zod schemas resolve correctly during the Vite build
- `Dockerfile.dev` (frontend + backend) — `CMD` now runs `pnpm install && pnpm dev` so the anonymous `node_modules` volume self-heals after adding new packages, eliminating the need for `docker compose down -v` when dependencies change
- `main.ts` — added `app.config.errorHandler` for visible error logging in dev

____

**Sources**

PR #248, opened 2026-03-28
Login and admin home page

### LINKED ISSUE #115: [DEV] Login page
**Location**

`/frontend/src/views/admin/Login.vue
____

**Description**

Implement the CMS login page. Accessible via a direct URL (e.g. /admin), without a visible login button on the public website. The page should authenticate the user against the backend login endpoint and store the session/token for subsequent CMS requests.

____

**Sources**

Q&A: "De publieke site hoeft geen login te hebben, die moet er vooral voor de admin kant zijn"

noahvsb:
**Issue(s)**

#115

____

**Description**

Added a login page, but in order to test to see if the currently logged in admin matches, I also made an admin home page. This page shows the username of the currently logged in admin, a profile picture (defaults to the favicon), a link to the CMS page and a logout button.

I also implemented the `checkUserIsAdmin` function that fetches the currently logged in user and stores it in a pinia store. This ensures that you don't need to fetch it twice (once to validate and another time to display the admin on a page). When the user isn't an admin, I redirect them to the login and pass a redirect URL parameter, which is used to redirect back after login.

Waiting on #223, so I can do the design.

(I forgot to commit during my work, so it's one huge commit)

**Update**

When doing the design, I also made a `useDarkMode` composable and an `AdminNavbar` for easy use in the views.

____

**Sources**

image
image
image

(it's beautiful :face_holding_back_tears:)


noahvsb:
#223 has been merged, so I can do the design. But I'm now waiting for #258 so I can incorporate the super admin into this PR.

frontend/src/components/admin/AdminNavbar.vue
```diff
@@ -0,0 +1,119 @@
...
+const router = useRouter();
+const route = useRoute();
+const { admin, clearAdmin } = useAuthStore();
+
+const currentLang = computed(() => i18n.global.locale.value as SupportedLang);
+const initials = computed(() => admin?.username?.slice(0, 2).toUpperCase() ?? "??");
+
+const profileOpen = ref(false);
+
+function handleClickOutside() {
```
lucasodb:
```suggestion
function handleClickOutside(e: MouseEvent) {
  const el = profileWrapper.value;
  if (el?.contains(e.target as Node)) return;
```

frontend/src/components/admin/AdminNavbar.vue
```diff
@@ -0,0 +1,119 @@
...
+const { t } = useI18n();
+const router = useRouter();
+const route = useRoute();
+const { admin, clearAdmin } = useAuthStore();
+
+const currentLang = computed(() => i18n.global.locale.value as SupportedLang);
+const initials = computed(() => admin?.username?.slice(0, 2).toUpperCase() ?? "??");
+
+const profileOpen = ref(false);
+
```
lucasodb, line 70:
```suggestion
const profileWrapper = ref<HTMLElement | null>(null);
```

frontend/src/components/NavControls.vue
```diff
@@ -0,0 +1,96 @@
...
+    </button>
+  </div>
+</template>
+
+<script setup lang="ts">
+import { ref, computed, onMounted, onUnmounted } from "vue";
+import { i18n, SUPPORTED_LANGS, type SupportedLang } from "@/i18n";
+import { useLocale } from "@/composables/useLocale";
+
+withDefaults(defineProps<{ isDark: boolean; variant?: "light" | "dark" }>(), {
```
lucasodb: "light" and "dark" are declared here, but not used in styles?

frontend/src/views/admin/AdminView.vue
```diff
@@ -0,0 +1,88 @@
...
+        <span v-if="admin?.super" class="super-badge">{{ t("admin.dashboard.superAdmin") }}</span>
+      </div>
+
+      <div class="section-divider" />
+
+      <!-- Quick actions -->
+      <section class="section">
+        <h2 class="section-title">{{ t("admin.dashboard.quickActions") }}</h2>
+        <div class="actions-grid">
+          <RouterLink :to="{ name: RouteNames.CMS }" class="action-card">
```
lucasodb: I think you should pass `params: { lang: currentLang }` here, like you do in `AdminNavbar`

lucasodb (changes requested):
When I click on the dropdown in `AdminNavbar.vue`, nothing happens, and I think I know why. The `handleClickOutside` handler always sets the menu to closed, it doesn't check whether the click was oustide of the dropdown, so as a result when clicking the dropdown `profileOpen` is set to true, but the same click also closes it again. So, the document click listener should check where the click came from instead of always setting the open value to false. The exact same thing is happening for the language menu for me, which is weird because I didn't notice earlier that this wasn't working. I guess this doesn't cause a problem for everyone, but since it can we should fix this. With the two additions I suggested this problem was fixed. You can do this for the language menu too, or make a seperate pr for this, since this isn't part of your pr.

frontend/src/components/admin/AdminNavbar.vue
noahvsb: I had that first, but the if is almost always true. I wanted to make a test where it wasn't for coverage, but I couldn't do it.

noahvsb: Ok, I just read your other comment, since it actually breaks things I'll change it back.

lucasodb (approved):

Issue #217, opened 2026-03-20
[DEV] Media Scraper


msth5:
**Location**

Backend

____

**Description**

Scrape the viernulvier live site for media from past productions. Needs to have a bulk and single id scrape variant. (Bulk for initial scrape, update to keep it up to date).

____

**Sources**

[API](https://www.viernulvier.gent/api/docs)

Issue #216, opened 2026-03-20
[DEV] Scraper Tags (and Tagtypes)


msth5:
**Location**

Backend

____

**Description**

Scrape the viernulvier live site for past tags. Needs to have a bulk and single id scrape variant. (Bulk for initial scrape, update to keep it up to date). Keep in mind the Tagtype.

____

**Sources**

[API](https://www.viernulvier.gent/api/docs)

Issue #215, opened 2026-03-20
[DEV] Scraper Hall and Pricing


msth5:
**Location**

Backend

____

**Description**

Scrape the viernulvier live site for past halls and pricing. Needs to have a bulk and single id variant. (Bulk for initial scrape, update to keep it up to date). This is an addition to the Event Scraper

____

**Sources**

[API](https://www.viernulvier.gent/api/docs)

Issue #214, opened 2026-03-20
[DEV] Scraper Events


msth5:
**Location**

Backend

____

**Description**

Scrape the viernulvier live site for past events. Needs to have a bulk and single id variant. (Bulk for initial scrape, update to keep it up to date)

____

**Sources**

[API](https://www.viernulvier.gent/api/docs)

PR #253, opened 2026-03-29
Feat/scraper helper

Acher0s:
**Issue(s)**

#214 #215 #216 #217
____

**Description**

Setup for the scraper, added an old_id field to several schemas to keep a link between objects in our system and the clients. Also an additional field "finalized" in producion, marking whether or not all events belonging to the production have been imported.

This PR includes changed to the DB schema, changes to TS schemas, changes in tests to handle those changes and test them accordingly.
____

**Sources**


msth5 (dismissed):
Nice baseline to start the other scrapers

backend/migrations/010.do.add_old_id.sql
noahvsb, line 1: There's already a 008 on the staging branch and Lucas made an 009 in #246. So would suggest calling this 010 and the next one 011.

noahvsb (changes requested):

backend/migrations/010.do.add_old_id.sql
noahvsb: Now the 009 is on staging.

Acher0s: Done

noahvsb (approved):
Changes look good 👍

PR #258, opened 2026-03-31
Feat/super admin

### LINKED ISSUE #256: [DEV] auth endpoint: add super admin
**Location**

`backend/migrations` we need an extra field in the database schema's
`backend/src/plugins/authorize.ts/jws.ts` to add a "super authentication" option
`backend/src/routes/auth/auth.ts` add the super authentication option to all admin things except for `/me`, `/login` and `/logout`.
`shared/src/types/admin.ts` add the super field
`DOCS/database(_nl).md` add to the EER

don't touch front-end, that will be fixed in #248, since there's already been progress there

____

**Description**

Add a "super" field to the admin that states an admin is a super admin or not. Only super admins are allowed to make fetch, create, edit and delete admins. Both are allowed to do all the other protected API requests.

____

**Sources**

noahvsb:
**Issue(s)**

closes #256 

____

**Description**

Implemented the super admin as stated in #256. I also found a small issue in the database markdowns and editted the create admin script.

____

**Sources**


noahvsb:
Will fix the coverage tomorrow, professor Dawyndt has us getting up early, so I'll be going to bed now.

shared/src/types/admin.ts
```diff
@@ -7,6 +7,7 @@ export const AdminBase = {
   id: primaryKey().describe("Primary key of admin."),
   username: z.string().max(32).describe("Username of admin."),
   profile_picture: z.url().max(2048).nullable().describe("Profile picture (url) of admin."),
+  super: z.boolean().describe("Whether admin has super privileges. (Being able to fetch, create, edit or delete other admins"),
```
lucasodb: You are missing a closing bracket here.

lucasodb (changes requested):
AdminSchema now requires the super field, but the auth handlers still only select/return id, username, and profile_picture. The responses are parsed with AdminSchema, so won't this validation fail? I believe every SELECT / RETURNING needs to include super. Other than that this looks fine to me, just one more nitpicky comment. Also, should we look into guarding against removing the last super admin? This is out of scope for now but might be interesting to look into in the future.

noahvsb:
> The responses are parsed with AdminSchema, so won't this validation fail?

It should, I'll look into this and I'll try to make it fail first and then fix it.

noahvsb:
> > The responses are parsed with AdminSchema, so won't this validation fail?
> 
> It should, I'll look into this and I'll try to make it fail first and then fix it.

I found it, the tests mock the SQL queries and return an admin with the super field. So there's no issue there, it's just that my SQL queries are lacking. In the future we should more properly test our SQL queries. I will make an issue for this.

noahvsb:
> I will make an issue for this.

#272

backend/src/routes/auth/handlers/create.ts
lucasodb:
```suggestion
 * @param request - The Fastify request, expected to contain `username`, `password`, and `super` in its body.
```

backend/src/routes/auth/handlers/replace.ts
```diff
@@ -5,17 +5,17 @@ import { buildQuery, getMetadata, parseParams, parseSchema } from "@/routes/help
...
 });
 
 const replaceAdminById = (server: FastifyInstance) =>
   buildQuery(
     server,
-    `UPDATE admin SET username = $1, password = $2, updated_by = $3, updated_at = $4
-     WHERE id = $5
+    `UPDATE admin SET username = $1, password = $2, super = $3, updated_by = $4, updated_at = $5
+     WHERE id = $6
      RETURNING id, username, profile_picture_url AS profile_picture`,
```
lucasodb:
```suggestion
     RETURNING id, username, super, profile_picture_url AS profile_picture`,
```

backend/src/routes/auth/handlers/replace.ts
lucasodb:
```suggestion
 * @param request - The Fastify request, expected to contain `username`, `password`, and `super` in its body.
```

backend/src/routes/auth/handlers/edit.ts
lucasodb:
```suggestion
 * @param request - The Fastify request, expected to contain a partial body with `username`, `password`, and/or `super`.
```

backend/src/routes/auth/handlers/edit.ts
```diff
@@ -35,12 +35,17 @@ export async function editAdmin(server: FastifyInstance, request: FastifyRequest
     values.push(await hashPassword(body.password));
   }
 
+  if (body.super !== undefined) {
+    fields.push(`super = $${i++}`);
+    values.push(body.super);
+  }
+
```
lucasodb, line 45: add this:

```js
if (fields.length === 0) {
  throw new HttpError(HttpClientError.BadRequest, "No fields to update"); 
}
```

backend/src/routes/auth/handlers/edit.ts
lucasodb: Write the parameters on a seperate line, for consistency.

backend/src/routes/auth/handlers/fetch.ts
lucasodb, line 109: Add a `_request` parameter.

lucasodb (changes requested):

backend/src/routes/auth/handlers/edit.ts
noahvsb: Simon already made a comment about this on the [authentication PR](https://github.com/SELab-2/viernulvier-1/pull/168#discussion_r2904493370). We should allow empty bodies since the [official specs](https://www.rfc-editor.org/rfc/rfc5789.html) don't state they should be disallowed.

noahvsb:
stupidest merge conflict ever

backend/src/routes/auth/handlers/fetch.ts
lucasodb, line 104: Sorry for being super nitpicky, but also add `@param _request ` here.

lucasodb (changes requested):

backend/src/routes/auth/handlers/fetch.ts
noahvsb: No worries, better safe than sorry.

lucasodb (approved):

PR #262, opened 2026-04-01
fix: remove .DS_Store from version control

### LINKED ISSUE #274: [BUG] legacy import test failing occasionaly
**Description**

Every so often I get `FAIL  test/legacy-import/shared.test.ts > parseLegacyImportCli > throws on unknown flag`.

See sources below for 2 examples where the tests fail.

It's not a high priority, because simply rerunning the validation fixes it.

____

**Environment**

github actions

**Sources**

https://github.com/SELab-2/viernulvier-1/actions/runs/23983112221/attempts/1?pr=248 

https://github.com/SELab-2/viernulvier-1/actions/runs/23955671580/attempts/1?pr=258

Simon-VdB:
**Issue(s)**

____

**Description**

macOS metadata file was accidentally committed to staging. The file is already covered by .gitignore so it will not be re-added in future commits.

____

**Sources**


noahvsb (approved):
Thx for also fixing that bug.

PR #263, opened 2026-04-01
make tag route use tag instead of tags

### LINKED ISSUE #218: [BUG] API routes sometimes use plural, sometimes singular
**Description**

For example you have `/api/v1/production` and `/api/v1/tags`. For consistency it's best to do only one and I think it's easier to use singular everywhere, since most routes have singular already.

____

**Environment**

**Sources**

lucasodb:
**Issue(s)**

Closes #218 

____

**Description**

Change tag endpoint route to the singular /api/v1/tag instead of /api/v1/tags, to be consistent with all other endpoints. 

____

**Sources**

### DIFF
--- a/backend/src/routes/tag/tags.ts
+++ b/backend/src/routes/tag/tags.ts
@@ -17,30 +17,30 @@ import {
  * Registers tag routes on the Fastify instance.
  *
  * @remarks
- * - `GET /api/v1/tags/all` — fetch tags (optionally filtered by production); includes non-public tags. 🔒
- * - `GET /api/v1/tags` — fetch public tags only (same query params as `/all`).
- * - `GET /api/v1/tags/:id/all` — fetch a single tag by ID (non-public allowed). 🔒
- * - `GET /api/v1/tags/:id` — fetch a single public tag by ID.
- * - `GET /api/v1/tags/:id/meta` — fetch a single tag with metadata by ID. 🔒
- * - `POST /api/v1/tags` — create a tag. 🔒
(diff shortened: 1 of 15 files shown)


backend/src/routes/tag/tags.ts
```diff
@@ -17,30 +17,30 @@ import {
...
- * - `GET /api/v1/tags/:id/all` — fetch a single tag by ID (non-public allowed). 🔒
- * - `GET /api/v1/tags/:id` — fetch a single public tag by ID.
- * - `GET /api/v1/tags/:id/meta` — fetch a single tag with metadata by ID. 🔒
- * - `POST /api/v1/tags` — create a tag. 🔒
- * - `PUT /api/v1/tags/:id` — replace a tag. 🔒
- * - `PATCH /api/v1/tags/:id` — partially update a tag. 🔒
- * - `DELETE /api/v1/tags/:id` — delete a tag by ID. 🔒
+ * - `GET /api/v1/tag/all` — fetch tags (optionally filtered by production); includes non-public tags. 🔒
+ * - `GET /api/v1/tag` — fetch public tags only (same query parameters as `GET /api/v1/tag/all`).
+ * - `GET /api/v1/tag/:id/all` — fetch a single tag by ID (non-public allowed). 🔒
```
noahvsb, line 22: **Not in the scope of this PR**

I feel like the /all should be gone from single tag fetches. It will be complicated to make the non-public tags still protected and the solution might not have the cleanest code, but the /all suffix feels wrong here.

noahvsb (approved):
Very straightforward change, looks good to me.

Issue #264, opened 2026-04-01
[DEV] Endpoints: blog post


Simon-VdB:
**Location**

`/backend/src/routes/blogpost/`

____

**Description**

Implement the blogs endpoint of the API.

/api/v1/blogpost/

____

**Sources**

PR #266, opened 2026-04-01
blogpost endpoint

Simon-VdB:
**Issue(s)**

https://github.com/SELab-2/viernulvier-1/issues/264

____

**Description**

Creates the blogpost endpoint for the backend.

____

**Sources**

N/A


shared/src/types/blog.ts
```diff
@@ -14,8 +14,10 @@ export type BlogWithMeta = z.infer<ReturnType<typeof BlogSchema.withMeta>>;
 
 export const BlogPostSchema = createSchema({
   id: primaryKey(),
-  content: z.string(),
   blog: foreignKey(() => BlogSchema),
+  title: z.string(),
+  content: z.record(z.string(), z.unknown()),
+  published_at: z.coerce.date().nullable(),
```
lucasodb, line 20: There should be a reference to `production` here.

backend/src/routes/blogpost/blogpost.ts
```diff
@@ -0,0 +1,29 @@
...
+ * - `GET /api/v1/blogpost/:id/meta` — fetch a single blogpost with metadata by ID. 🔒
+ * - `POST /api/v1/blogpost` — create a new blogpost. 🔒
+ * - `PUT /api/v1/blogpost/:id` — replace an existing blogpost's data. 🔒
+ * - `PATCH /api/v1/blogpost/:id` — partially update an existing blogpost. 🔒
+ * - `DELETE /api/v1/blogpost/:id` — delete a blogpost by ID. 🔒
+ *
+ * @param server - The Fastify instance to register routes on.
+ */
+export default function blogPostRoutes(server: FastifyInstance) {
+  const protect = { preHandler: [server.authorize] };
```
lucasodb: In #258 authorize gets an optional argument for the admin status, so if that pr is merged first than you'll need to make this server.authorize().

backend/src/routes/blogpost/handlers/create.ts
```diff
@@ -0,0 +1,50 @@
...
+    `INSERT INTO blogpost (blog, title, content, published_at, created_by, updated_by, created_at, updated_at)
+     VALUES ($1, $2, $3, $4, $5, $5, $6, $6)
+     RETURNING id, blog, title, content, published_at`,
+    z.tuple([
+      CreateBlogPostBodySchema.shape.blog,    // blog (FK)
+      CreateBlogPostBodySchema.shape.title,   // title
+      CreateBlogPostBodySchema.shape.content, // content (JSONB)
+      CreateBlogPostBodySchema.shape.published_at, // published_at
+      z.int(),                                // admin
+      z.date(),                               // current_time
```
lucasodb: Leave out these inline comments for consistency.

backend/src/routes/blogpost/handlers/replace.ts
```diff
@@ -0,0 +1,55 @@
...
+     WHERE id = $7
+     RETURNING id, blog, title, content, published_at`,
+    z.tuple([
+      ReplaceBlogPostBodySchema.shape.blog,    // blog (FK)
+      ReplaceBlogPostBodySchema.shape.title,   // title
+      ReplaceBlogPostBodySchema.shape.content, // content (JSONB)
+      ReplaceBlogPostBodySchema.shape.published_at, // published_at
+      z.int(),                                 // admin
+      z.date(),                                // current_time
+      z.int(),                                 // id
```
lucasodb: Same here.

shared/src/types/blog.ts
```diff
@@ -14,8 +14,10 @@ export type BlogWithMeta = z.infer<ReturnType<typeof BlogSchema.withMeta>>;
 
 export const BlogPostSchema = createSchema({
   id: primaryKey(),
-  content: z.string(),
   blog: foreignKey(() => BlogSchema),
+  title: z.string(),
```
lucasodb: `title` is a string in the schema, but VARCHAR(255) in the DB. So, it's best to add `.max(255)` for consistency.

lucasodb (changes requested):
Mostly looks good. As we discussed, many blogposts can and initially will reference the same blog, but currently in the migrations there is a `UNIQUE` on the blog_id FK, which implies at most one post per blog. So, make a new migrations that drops the UNIQUE. 
Also, I understand that we want to use NULL as the value for `published_at` for the draft state of a blogpost, but GET on the blogpost endpoint returns all blogposts. I assume we don't want drafts to be public, so maybe add something like `WHERE published_at IS NOT NULL` to the fetch handler to make sure only published productions are visible (by non-admins).

noahvsb (changes requested):
After #278 has been merged, I would like you to also add blogpost SQL integration tests. If your SQL is already functional, all you need is 1 prompt.

lucasodb (commented):
All changes I requested were correcly implemented. I won't approve yet, since Noah requested that you also write integration tests, but apart from that this is all good.

backend/src/routes/blogpost/blogpost.ts
```diff
@@ -0,0 +1,29 @@
...
+export default function blogPostRoutes(server: FastifyInstance) {
+  const protect = { preHandler: [server.authorize()] };
+
+  server.get("/api/v1/blogpost", replyHandler(server, fetchBlogPosts));
+  server.get("/api/v1/blogpost/:id", replyHandler(server, fetchBlogPost));
+  server.get("/api/v1/blogpost/:id/meta", protect, replyHandler(server, fetchBlogPostWithMeta));
+  server.post("/api/v1/blogpost", protect, replyHandler(server, createBlogPost));
+  server.put("/api/v1/blogpost/:id", protect, replyHandler(server, replaceBlogPost));
+  server.patch("/api/v1/blogpost/:id", protect, replyHandler(server, editBlogPost));
+  server.delete("/api/v1/blogpost/:id", protect, replyHandler(server, deleteBlogPost));
```
noahvsb: Maybe use `/api/v1/blog/post`?

backend/test/sql.integration.test.ts
```diff
@@ -342,6 +342,151 @@ describe("Blog routes — SQL integration", { sequential: true }, () => {
   });
 });
 
+describe("BlogPost routes — SQL integration", { sequential: true }, () => {
+  let blogId: number;
+  let blogPostId: number;
+
+  test("POST /api/v1/blog — creates a blog for blogpost tests", async () => {
```
noahvsb: This should just be a `beforeAll` instead of a test.

backend/test/sql.integration.test.ts
```diff
@@ -342,6 +342,151 @@ describe("Blog routes — SQL integration", { sequential: true }, () => {
...
+      url: `/api/v1/blogpost/${blogPostId}`,
+      cookies: { session: sessionCookie },
+    });
+    expect(deleteResponse.statusCode).toBe(HttpSuccess.OK);
+
+    const fetchResponse = await server.inject({ method: "GET", url: `/api/v1/blogpost/${blogPostId}` });
+    expect(fetchResponse.statusCode).not.toBe(HttpSuccess.OK);
+  });
+
+  test("DELETE /api/v1/blog/:id — cleans up the blog used for blogpost tests", async () => {
```
noahvsb: This should be an `afterAll`.

noahvsb (changes requested):

noahvsb (approved):
Requested changes look good to me 👍

lucasodb (approved):

Issue #112, opened 2026-02-26
[DEV] Production detail page


Simon-VdB:
**Location**

`/frontend/src/views/ProductionDetail.vue`

____

**Description**

Implement the production detail page that displays all information about a single production. This includes: hero image (FE3_header crop), supertitle, artist name, production title, tags as pills, description text, practical info box, list of linked events (with dates, times, hall), and a link to the series/tag page if the production is part of a series (e.g. "Onderdeel van → HOUSEWARMING"). Content must be displayed in the selected language.

____

**Sources**
- Current viernulvier.gent production detail page
- Kickoff document slide 6: "productie detail"

PR #273, opened 2026-04-03
Feat/production detail

trherber:
**Issue(s)**

#112 

____

**Description**

This PR implements the **Production Detail View**.

It displays a **Hero** with an **optional photo** and the most important overview data, such as **title, artist, supertitle, tagline, performance period, duration, and genre**. 

Below that is the **Detail Section**, which includes:
* Optional **quote and quote source**.
* The **description** and **description_2** (displayed in a slightly different color).
* A **sidebar** with collapsible **tags** and a box containing a **teaser and description_extra**.
* An **extra info section** at the bottom of the description where **info and program** data can be displayed.

Following this is an **Events Section** that shows a collapsible list of all events with their **date, start/end times, hall, address, and price range**. 

Finally, there are **Gallery** and **Blogpost** sections. These are currently placeholders for future PRs, but the design is already implemented and can be used as a foundation to build upon.

I also built a **Not Found** view and component that allows for custom message props.

### Limitations
* **Banner Image:** The banner currently uses a hardcoded image and still needs to be replaced with the correct media, but this also seems easier to do once we have the data.
* **Feedback:** A lot of things might still need to change, but I’d rather open the PR now so I can work with specific feedback instead of spending an unnecessary amount of time on it by myself.

### UI / Preview
I have included a screenshot below so you can see where everything is displayed, as these fields are not present in the legacy data and are therefore hidden in the UI. To show the positioning, I have copied some random text from the API into the detail component. (Please disregard the actual content, it is only used to visualize the fields that would otherwise be hidden.)

image
____

**Sources**


frontend/src/components/production/HeroSection.vue
```diff
@@ -0,0 +1,131 @@
...
+        alt="Contemporary dance performance"
+        class="h-full w-full object-cover grayscale contrast-125"
+        src="https://lh3.googleusercontent.com/aida-public/AB6AXuDU21ewyTmp0aPBEaU9xMGtqD-vLtugWzQIgeF5cNgkw3uvYnZCH2696RoNKJ-kPMsE9zWgEZnutEMH-J-JJfxTKGmKYwRmPRlTakVaNy15qZ9KQXigCvuBMADoZpkjjd9YYLpjteDwWRYyjupc87R0EJsRbW2MzKJN0p2HzFbO65oUEoM7xAE0CfFOdMkaUI45I3xa5PPVHFIdzJkm4Mtz8Y1xX_3ALsoeMsj0C_QTlHCfJnFU0vs_2s95CJhFPdbRZSLDLIPgysph"
+        referrerPolicy="no-referrer"
+      />
+      <div class="absolute inset-0 bg-linear-to-t from-black/80 via-black/20 to-transparent"></div>
+    </div>
+    
+    <div class="relative z-10 mx-auto flex w-full max-w-7xl flex-col justify-between gap-12 px-6 pb-20 md:flex-row md:items-end md:px-12">
+      <div ref="heroContent" class="opacity-0 animate-fade-up max-w-4xl">
```
lucasodb: `heroContent` is unused, not sure if you still plan to do something with this.

frontend/src/components/production/HeroSection.vue
```diff
@@ -0,0 +1,131 @@
...
+          <h1 class="text-[clamp(2.5rem,8vw,8rem)] font-black uppercase leading-[0.9] tracking-tighter text-ink-on-inv wrap-break-word">
+            {{ content.title }}
+          </h1>
+          <p class="italic text-xl font-light leading-tight text-ink-on-inv opacity-90 md:text-2xl">
+            {{ content.tagline }}
+          </p>
+        </div>
+      </div>
+
+      <div ref="statsContent" class="opacity-0 animate-fade-in flex flex-col gap-6 text-ink-on-inv md:items-end">
```
lucasodb: `statsContent` is also unused.

frontend/src/composables/useProductionEvents.ts
```diff
@@ -0,0 +1,66 @@
...
+  maxPrice: number | null;
+}
+
+/**
+ * Fetches events for a production and attaches hall info and prices.
+ */
+export function useProductionEvents(productionId: number) {
+  const events = ref<Event[]>([]);
+  const halls = ref<Hall[]>([]);
+  const loading = ref(true);
```
lucasodb, line 40: `loading` isn't actually used here yet, so it's best to wire it, so there aren't any unexpected layout shifts etc.

frontend/src/composables/useProductionEvents.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  onMounted(async () => {
+    try {
```
lucasodb, line 47: You only do try/finally here, best to add catch + error, like you to in `useTagGroups`.

frontend/src/composables/useTagGroups.ts
```diff
@@ -0,0 +1,88 @@
...
+    for (const tag of fetchedTags.value) {
+      const typeId = tag.tag_type as number;
+      const translatedName = tProd(tag.name).trim();
+
+      if (translatedName) {
+        if (!grouped.has(typeId)) {
+          grouped.set(typeId, []);
+        }
+
+        grouped.get(typeId)!.push(tProd(tag.name));
```
lucasodb: You compute `const translatedName = tProd(tag.name).trim()`, but then push `tProd(tag.name)` again into the group. I think you meant to push `translatedName` here instead?

lucasodb (changes requested):
This looks really good already, great job. I've tested this with a bunch of real API data, and mostly everything looks good. I've just left a few small comments. I would also already start with parsing the html, this is actually really easy and only a small change. Also, the legacy csv data uses very peculiar line endings, so the descriptions for those productions are full of backslashes, so I would parse these values to make them look better.

lucasodb:
Great. Just some small annoying issues to deal with left.

As you can see from the image, the backslashes aren't fully parsed away yet. It looks like you missed the <not \> \ <newline> case. Also, I think we should try to lower the amount of newlines in the description fields. This is not just in the legacy data, the api data has this too. I don't know why, but sometimes they have like 5 empty lines between text blocks, which looks a bit weird, so you could try to make it so that max 2 or 3 empty lines between texts are shown.

Other than that, there is still the problem of relative URLs, for both images and just pages, but I think you already mentioned this. This is kind of difficult to deal with and maybe something to think about for milestone 3. 

There is just 1 more thing that still bothers me a little, and that is that the detail section block still shows up, even though there is no data to show here (like in the second image). I think it would be better to leave this out, if nothing is shown there.

image

image

trherber:
Maybe for Milestone 3 we should show all ticket prices instead of a range. Since there are often tickets for €0, the range usually starts at zero and doesn't add much info. Is it okay if I leave it as is for now? We can ask the client how important the price display is to them.

frontend/src/utils/parsers.ts
```diff
@@ -6,6 +6,8 @@ function normalize(input: string): string {
     .replace(/\\n/g, "\n")
     .replace(/\r\n/g, "\n")
     .replace(/^\s*\\\s*$/gm, "")
+    .replace(/\n{3,}/g, "\n\n")
```
lucasodb: I would just replace this by `.replace(/[\r\n]{3,}/g, "\n\n")`, so all line brakes are matched.

lucasodb (changes requested):
Just this tiny change, then I think this can be merged, all the other minor things can be solved later.

lucasodb (approved):

PR #278, opened 2026-04-05
SQL integration tests

### LINKED ISSUE #272: [DEV] integration tests: test SQL queries
**Location**

`backend/test/routes`

____

**Description**

Currently the `server.pg.query` function is mocked in all of our tests, which means we don't actually test the SQL queries, which is fine and actually wanted for unit tests. But we don't have any integration tests to actually test the SQL. My suggestion is to use `PostgreSqlContainer` from the [`testcontainers` library](https://node.testcontainers.org/).
____

**Sources**

noahvsb:
**Issue(s)**

#272 
____

**Description**

Testing the SQL queries used in all of the endpoint routes.

- [x] hall
- [x] auth
- [x] blog
- [x] production
- [x] event
- [x] event price
- [x] tag
- [x] tag type

____

**Sources**

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -43,6 +43,7 @@
   "description": "",
   "devDependencies": {
     "@eslint/js": "^10",
+    "@testcontainers/postgresql": "^11.13.0",
     "@types/bcrypt": "^6.0.0",
     "@types/node": "^25.5.0",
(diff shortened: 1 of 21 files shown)


msth5 (approved):
assuming most changes came as result from integration tests, seems to be nothing major

Issue #279, opened 2026-04-05
[DEV] text editor component


Simon-VdB:
**Location**

/frontend/src/components/TextEditor.vue
____

**Description**

Create a reusable vue component for a text editor.

____

**Sources**

https://simplemde.com

PR #280, opened 2026-04-06
markdowneditor component

Simon-VdB:
**Issue(s)**

(https://github.com/SELab-2/viernulvier-1/issues/279)

____

**Description**

made a reusable vue component for text editing.

____

**Sources**

https://simplemde.com


noahvsb (changes requested):
There are no tests and also no way to manually test the editor yet. I would just add this editor to the admin home page temporarily so we can test it there.

frontend/test/views/HomeView.test.ts
```diff
@@ -75,4 +89,56 @@ describe("HomeView.vue", () => {
...
+
+  describe("markdown editor preview", () => {
+    it("updates previewContent when MarkdownEditor emits update:modelValue", async () => {
+      const editor = wrapper.findComponent(MarkdownEditor);
+      await editor.vm.$emit("update:modelValue", "hello");
+      // No error thrown — the v-model setter is exercised
+    });
+  });
+
+  // ── Dark mode — logic owned by HomeView ──────────────────────────────────
```
noahvsb: This is not logic owned by HomeView.

frontend/src/components/MarkdownEditor.vue
```diff
@@ -0,0 +1,158 @@
...
+const props = defineProps<{
+  modelValue: string;
+  placeholder?: string;
+}>();
+
+const emit = defineEmits<{
+  "update:modelValue": [value: string];
+}>();
+
+const textarea = ref<HTMLTextAreaElement | null>(null);
```
noahvsb: fix warning

frontend/src/components/MarkdownEditor.vue
```diff
@@ -0,0 +1,158 @@
...
+let editor: EasyMDE | null = null;
+
+onMounted(() => {
+  /* v8 ignore next -- Vue guarantees the ref is populated before onMounted */
+  if (!textarea.value) return;
```
noahvsb: If vue guarantees this, then you can just leave out this line I would think.

frontend/src/components/MarkdownEditor.vue
```diff
@@ -0,0 +1,158 @@
...
+  });
+
+  editor.codemirror.on("change", () => {
+    if (editor) emit("update:modelValue", editor.value());
+  });
+});
+
+onUnmounted(() => {
+  /* v8 ignore next -- editor is always set when onUnmounted runs after a successful mount */
+  editor?.toTextArea();
```
noahvsb: You can also leave out this line if it's always set.

noahvsb (changes requested):
Image

It works fine and looks good, but not in dark mode and only on its own. I can hardly see the cursor for example and it feels like it doesn't blend in well with our current style.

frontend/src/components/MarkdownEditor.vue
```diff
@@ -18,13 +18,10 @@ const emit = defineEmits<{
   "update:modelValue": [value: string];
 }>();
 
-const textarea = ref<HTMLTextAreaElement | null>(null);
+const textarea = ref<HTMLTextAreaElement>(null!);
```
noahvsb: The warning is still there.

noahvsb (changes requested):

frontend/src/components/MarkdownEditor.vue
Simon-VdB: what is the warning because I don't see it.

noahvsb: It's a linter warning, I suggest installing the eslint extension for vscode so you can see them. You can also fix it by running `pnpm run lint:fix`.

noahvsb (approved):
Looks good now 👍

Issue #277, opened 2026-04-05
[DEV] fix the bump typescript from 5.9.3 to 6.0.2 PR


noahvsb:
**Location**

#268 
____

**Description**

Dependabot wants to merge #268, but since it's a major version upgrade, stuff is breaking.

I would suggest merging #268 in a `fix/*` branch and fixing it on that branch instead of directly on the dependabot branch.
____

**Sources**

PR #282, opened 2026-04-06
Fix typescript bump

Simon-VdB:
**Issue(s)**

https://github.com/SELab-2/viernulvier-1/issues/277

____

**Description**

Bumps TypeScript from 5.9.3 to 6.0.2 and updates all three packages to be compatible with the new compiler version. The main breaking changes addressed are the requirement to explicitly set rootDir when outDir is configured (TS5011), and the deprecation of baseUrl (TS5101) — the latter is resolved cleanly by removing baseUrl entirely and making all paths entries explicitly relative, which has been supported since TypeScript 5.0. A yargs .locale("en") call was added to legacy-import/shared.ts to prevent test failures on systems with a non-English locale, and a bare module import in scripts/scraper/scrape.ts that relied on baseUrl resolution was updated to a proper relative import.

____

**Sources**

N/A


noahvsb (approved):
Looks good to me, validation succeeds.

PR #283, opened 2026-04-06
validate legacy data before inserting

### LINKED ISSUE #275: [DEV] legacy import: do zod parsing on tags, events and productions before inserting
**Location**

`backend/test/legacy-import`

____

**Description**

The legacy import worked fine before, but the zod schema's got changed for the scraper and now they fail. However the tests don't catch that, because the zod schema's aren't used to parse anything in the legacy imports. So it's best to parse the objects before inserting them.

I was also thinking about maybe doing the legacy imports with API calls instead of direct SQL queries, but that will introduce 2 issues:

1. On what admin should we login to authenticate? Maybe we should have a token for this sort of stuff.
2. The extra call to the API for each insert (since we don't have a bulk insert (yet)) will make it way slower.

____

**Sources**

lucasodb:
**Issue(s)**

Closes #275 

____

**Description**

Legacy importers now validate the data with the Zod schemas before inserting into the DB.

____

**Sources**

### DIFF
--- a/backend/src/legacy-import/import-events-legacy.ts
+++ b/backend/src/legacy-import/import-events-legacy.ts
@@ -11,6 +11,13 @@ import {
   type CsvRecord,
   type ImportArgs,
 } from "./shared.js";
+import {
+  formatLegacyZodError,
+  legacyEventCreateBody,
+  legacyHallInsertBody,
+  LegacyHallInsertSchema,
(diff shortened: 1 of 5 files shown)


noahvsb (approved):
I tested it locally and all zod validation succeeds, well done. The code style also looks good.

PR #287, opened 2026-04-07
Feat/CMS visualising data

willemDC-Info:
**Issue(s)**

____

**Description**

This PR introduces a grid view implementation for productions in the CMS to visualise all data currently in the database for productions and there events and tags.

The grid view is built using AG Grid and includes filtering, column selection, and support for displaying larger text fields. Users can now customize which columns are visible and easily search/filter through productions.
____

**Sources**

AG Grid documentation: https://www.ag-grid.com/


frontend/src/services/api.ts
noahvsb, line 1: These changes are not necessary, it works perfectly fine with cookies. I also checked the usages and `getStoredAuthToken()` is never used, not even in tests.

frontend/src/services/auth.ts
noahvsb, line 1: These changes are also unnecessary.

frontend/src/assets/stylesheets/cms-ag-grid.css
noahvsb, line 1: Move this to `frontend/src/assets/stylesheets/`

frontend/src/services/cms/helpers.ts
noahvsb, line 1: Move this to the `services` folder and then you can just call it `cms.ts`.

frontend/src/style.css
```diff
@@ -43,4 +46,115 @@
   #app {
     @apply min-h-screen w-full;
   }
+
```
noahvsb, line 63: Add a comment to make this section more distinct from the previous one.

frontend/src/views/admin/CMSView.vue
```diff
@@ -1,13 +1,1665 @@
 <template>
-  <div>
-    <AdminNavbar :is-dark="isDark" @toggle-dark="toggleDark" />
-    <span>CMS (admin only)</span>
+  <div class="min-h-screen bg-surface-0">
+    <AppNavbar :is-dark="isDark" @toggle-dark="isDark = !isDark" />
```
noahvsb:
```suggestion
    <AdminNavbar :is-dark="isDark" @toggle-dark="toggleDark />
```

frontend/src/views/admin/CMSView.vue
```diff
@@ -1,13 +1,1665 @@
...
+          />
+
+          <div class="flex items-center gap-2">
+            <span class="cms-selected-chip">
+              {{ t("cms.actions.selectedCount", { count: selectedCount }) }}
+            </span>
+          </div>
+        </div>
+
+        <div class="cms-grid-actions">
```
noahvsb: Add locale for this, currently it's always in english.

frontend/src/views/admin/CMSView.vue
```diff
@@ -1,13 +1,1665 @@
...
+  teaser: Record<SupportedLang, string>;
+  supertitle: Record<SupportedLang, string>;
+  description: Record<SupportedLang, string>;
+  description_2: Record<SupportedLang, string>;
+  video_1: Record<SupportedLang, string>;
+  video_2: Record<SupportedLang, string>;
+}
+
+const { t } = useI18n();
+const isDark = ref(getInitialDark());
```
noahvsb:
```suggestion
const { isDark, toggleDark } = useDarkMode();
```

frontend/src/views/admin/CMSView.vue
```diff
@@ -1,13 +1,1665 @@
...
+          </button>
+        </footer>
+      </section>
+    </div>
   </div>
 </template>
 
 <script setup lang="ts">
-import { useDarkMode } from "@/composables/useDarkMode";
-import AdminNavbar from "@/components/admin/AdminNavbar.vue";
```
noahvsb, line 10: Keep these

frontend/src/views/admin/CMSView.vue
noahvsb, line 1: Try to split this up more, have some different components like the `HomeView.vue` and also try to move more functions to the `cms-helpers.ts` and maybe even split that file up if it gets too big. (I would do `frontend/src/services/cms/xxx.ts` if you think you need to split it up.)

Currently I haven't really reviewed most of the content of this file, because I want to see it more split up first.

frontend/src/styles/cms-ag-grid.css
```diff
@@ -0,0 +1,245 @@
+.cms-grid-shell {
+  position: relative;
+  overflow: hidden;
+  border: 1px solid var(--surface-3);
+  border-radius: 1.5rem;
```
noahvsb: You should lower this to almost zero, it feels a bit too rounded and crops off some things. It should be very subtly rounded.

frontend/test/main.test.ts
noahvsb, line 1: Because of these changes App.vue has 0% coverage, and it shows up red in the coverage results. I don't know how the vitest coverage isn't failing, because it should be 80% coverage.

noahvsb (changes requested):
I like that you made a composable, nicely split up.

I have a few issues with the tag part, but that's out of scope.

frontend/src/views/admin/CMSView.vue
noahvsb: I would also make this a component if possible.

frontend/src/style.css
noahvsb: @willemDC-Info you forgot this requested change.

frontend/src/views/admin/CMSView.vue
noahvsb: Add all these types and interfaces to `services/cms/types.ts`.

frontend/src/views/admin/CMSView.vue
```diff
@@ -459,7 +330,7 @@ import {
   makeEditorValues,
   toIsoStringFromLocalInput,
   toLocalDateTimeInput,
-} from "./cms-helpers";
+} from "@/services/cms";
 
 type EventGridRow = CmsEventGridRow;
 type ProductionGridRow = CmsProductionGridRow;
```
noahvsb: Remove these two typing aliases.

noahvsb (changes requested):
Mostly looks good already.

For the scripts, I feel like they should be more split up, but for now it's good enough. However in the future if we can to reuse them for other views of the CMS, like tag or admin management, we should have them split up so they can be reused.

frontend/src/styles/cms-ag-grid.css
noahvsb: @willemDC-Info you also forgot this requested change.

frontend/frontend/src/services/cms/types.ts
noahvsb, line 1: frontend/frontend? remove this file

noahvsb (changes requested):

noahvsb (approved):
Changes are correctly implement, looks good to me now.

PR #288, opened 2026-04-07
fix legacy import path for windows

trherber:
**Issue(s)**

____

**Description**

Fix duplicate c:/c:/ in legacy import paths on Windows.

____

**Sources**


noahvsb (approved):
It works on linux as well.

PR #294, opened 2026-04-07
frontend service for fetching tags by production id

trherber:
**Issue(s)**

____

**Description**

Updated frontend tags service to sync with the backend query-string filtering (?production=ID).
(Added an optional productionId parameter to getTags and getAllTags services)

____

**Sources**

### DIFF
--- a/frontend/src/services/tags.ts
+++ b/frontend/src/services/tags.ts
@@ -89,7 +89,7 @@ export type UpdateTagTypeInput = Partial<CreateTagTypeInput>;
 /**
  * Fetches publicly visible tags. 
  * Can be filtered by production ID.
- * * @param productionId - Optional ID to get tags for a specific production.
+ * @param productionId - Optional ID to get tags for a specific production.
  * @example
  * const tags = await getTags(45); // Get public tags for production 45
  */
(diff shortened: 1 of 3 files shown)


frontend/src/services/tags.ts
```diff
@@ -113,15 +113,13 @@ export async function getTag(id: number): Promise<Tag> {
 // ---------------------------------------------------------------------------
 
 /**
- * Fetches all tags, including hidden ones (public: false).
- *
- * @throws {ApiError} 401 — unauthenticated.
```
lucasodb, line 118: I would keep this, because /tag/all is still behind authorize(), and otherwise it's not clear to callers that you still need a session for this route.

frontend/src/services/tags.ts
```diff
@@ -87,15 +87,15 @@ export type UpdateTagTypeInput = Partial<CreateTagTypeInput>;
 // ---------------------------------------------------------------------------
 
 /**
- * Fetches all publicly visible tags (public — no session required).
- *
- * Only tags with `public: true` are returned.
- *
+ * Fetches publicly visible tags. 
+ * Can be filtered by production ID.
+ * * @param productionId - Optional ID to get tags for a specific production.
```
lucasodb: There is an extra * here.

frontend/src/services/tags.ts
```diff
@@ -113,15 +113,13 @@ export async function getTag(id: number): Promise<Tag> {
...
 /**
- * Fetches all tags, including hidden ones (public: false).
- *
- * @throws {ApiError} 401 — unauthenticated.
- *
- * @example
- * const allTags = await getAllTags();
+ * Fetches all tags (including hidden ones).
+ * Can be filtered by production ID.
+ * * @param productionId - Optional ID to get all tags for a specific production.
```
lucasodb: Here too.

lucasodb (changes requested):
Looks good, mirrors the pattern you used for events. Just two very minor comments.

lucasodb (approved):

PR #293, opened 2026-04-07
productions page basic implementation

lucasodb:
**Issue(s)**

Closes #289 

____

**Description**

Basic productions page, that fetches all productions. The production cards display the title, artist, date(s), hall(s) and tag(s). There is no pagination, filters, etc. implemented yet.

____

**Sources**

PR #295, opened 2026-04-07
add pagination to productions page

### LINKED ISSUE #291: [DEV] pagination for productions page
**Location**

`frontend/src/utils`, `frontend/src/views/ProductionsView.vue`

____

**Description**

Create pagination for the productions page, so only a limited amount of productions are fetched at a time.

____

**Sources**

lucasodb:
**Issue(s)**

Closes #291 

____

**Description**

Implemented pagination for the productions page.

API (GET /production):
- Responses are now always { items, total }: items is the ordered list of productions, total is the number of rows in the database (not just the current page).
- Optional query parameter limit and offset added
- Without limit: behavior matches a full list fetch: all productions in items, and total === items.length.
- With limit: one page of rows is returned and total is the full count so clients can build pagination.

Frontend:
- getProductions in productions helper accepts an optional { limit?, offset? } object and builds the query string accordingly, leaving out options keeps a single request for the full list.
- ProductionsView loads the public list in pages (of 20), shows prev/next and a page input, and displays range when there is more than one page.

Going back to the top of the page when navigating to a different page doesn't feel very smooth yet, so I'll make an issue to try and improve this in the future.

____

**Sources**

### DIFF
--- a/backend/src/routes/production/handlers/fetch.ts
+++ b/backend/src/routes/production/handlers/fetch.ts
@@ -1,4 +1,5 @@
 import type { FastifyInstance, FastifyRequest } from "fastify";
+import type { QueryResult } from "pg";
 import type {ProductionWithBackwardsRefs, ProductionWithMeta} from "@viernulvier/shared/index.js";
 import {ProductionSchema, ProductionSchemaWithBackwardsRefs, stringToInt} from "@viernulvier/shared/index.js";
 import { parseParams, parseSchema, ParseContext } from "@/routes/helpers.js";
@@ -153,6 +154,7 @@ export type PaginatedProductions = {
  *
  * @param server - The Fastify instance, used for database access and logging.
  * @param request - The Fastify request; optional `limit` and `offset` query params.
+ * @returns The list of productions as `{ items, total }`; throws if parsing failed.
  */
 export async function fetchProductions(
   server: FastifyInstance,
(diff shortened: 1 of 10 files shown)


backend/src/routes/production/handlers/fetch.ts
```diff
@@ -128,21 +128,73 @@ export async function fetchProductionWithMeta(
...
+    `${ProductionSelect} ORDER BY p.id ASC LIMIT $1 OFFSET $2`,
+    [limit, offset],
+  );
+  const items = parseSchema(
+    server,
+    z.array(ProductionSchemaWithBackwardsRefs),
+    result.rows,
+    ParseContext.Database,
+  );
+  return { items, total };
```
noahvsb, line 197: I would just set the result and total variable in the if-else-statement and then you can share the items and return code for both cases for less code duplication.

backend/src/routes/production/handlers/fetch.ts
```diff
@@ -128,21 +128,73 @@ export async function fetchProductionWithMeta(
...
+
 /**
  * Fetches a list of productions.
  *
+ * - Without `limit`: returns every production (same ordering as before), as `{ items, total }`.
+ * - With `limit`: returns a page `{ items, total }` where `total` is the full row count.
+ *
  * @param server - The Fastify instance, used for database access and logging.
- * @param _request - The Fastify request (currently unused, reserved for future filters).
- * @returns The list of productions, or `null` if parsing failed.
```
noahvsb, line 136: Don't remove the @returns.

frontend/src/services/productions.ts
```diff
@@ -81,16 +81,35 @@ export interface BulkUpdateProductionsInput {
...
+  items: ProductionWithBackwardsRefs[];
+  /** Total number of productions (all pages), not just `items.length`. */
+  total: number;
+};
+
 /**
- * Fetches all productions (public — no session required).
+ * Fetches productions (public — no session required).
  *
- * @returns Array of productions, each with `tags` and `events` as arrays of linked IDs.
```
noahvsb, line 87: Here also don't remove the @returns.

frontend/src/views/ProductionsView.vue
noahvsb, line 36: Out of scope for this PR, but it would be nice to have one of those placeholder loading animations instead of some text. Can you make an issue for this? I would argue it's not needed for milestone 2, so you can set it to milestone 3.

frontend/src/views/ProductionsView.vue
```diff
@@ -86,8 +156,32 @@ import {
...
+    doc.scrollTop = 0;
+  }
+  if (typeof body.scrollTo === "function") {
+    body.scrollTo({ top: 0, left: 0, behavior: "smooth" });
+  } else {
+    body.scrollTop = 0;
+  }
+}
+
 function getInitialDark(): boolean {
```
noahvsb: I missed it on #293, but you should use the `useDarkMode` composable for dark mode, it makes the dark mode implementation much cleaner and easier.

noahvsb (changes requested):
To make our front-end stateless, you should include a query parameter for the page index. Currently when I'm on page 156 for example and I refresh, I go back to page 1.

I like the page navigation buttons at the top, it's nice to immediately see the page info without having to scroll all the way down. Plus that way you can also spam the next page button without having to scroll all the way down between each click.

lucasodb:
> To make our front-end stateless, you should include a query parameter for the page index. Currently when I'm on page 156 for example and I refresh, I go back to page 1.
> 
> I like the page navigation buttons at the top, it's nice to immediately see the page info without having to scroll all the way down. Plus that way you can also spam the next page button without having to scroll all the way down between each click.

I don't really think that you would be spamming the next page button. You would never really go to all productions, even when filtering on a certain tag for example. Mostly you will just search for a specific production, and you wouldn't just go through the list of productions. I prefer the navigation buttons at the bottom, like on the original site. I could make it so the page doesn't scroll to the top every time, that way you could still spam click the buttons. But again, for the type of usage we expect this isn't really necessary, and buttons at the top would also like a bit cluttered I think, with all the filter and search options being there too.

noahvsb:
> I don't really think that you would be spamming the next page button. You would never really go to all productions, even when filtering on a certain tag for example. Mostly you will just search for a specific production, and you wouldn't just go through the list of productions. I prefer the navigation buttons at the bottom, like on the original site. I could make it so the page doesn't scroll to the top every time, that way you could still spam click the buttons. But again, for the type of usage we expect this isn't really necessary, and buttons at the top would also like a bit cluttered I think, with all the filter and search options being there too.

Alright, then keep it at the bottom. Me and the 5 other top page controls enjoyers will be disappointed 😔
.

noahvsb (approved):
The way the changes are implemented looks good to me.

PR #300, opened 2026-04-09
add search features to productions page

### LINKED ISSUE #298: [DEV] search feature for productions page
**Location**

`frontend/src/views/ProductionsView.vue`

____

**Description**

Implement a search bar on the productions page, that let's you search on title, artist, teaser, tagline, description and hall name.

____

**Sources**

lucasodb:
**Issue(s)**

Closes #298 

____

**Description**

Added a search bar at the top of the productions page, which fetches productions who's title, artist, tagline, teaser, description or who's event's hall name(s) includes the search query.

____

**Sources**

### DIFF
--- a/backend/src/routes/production/handlers/fetch.ts
+++ b/backend/src/routes/production/handlers/fetch.ts
@@ -131,12 +131,36 @@ export async function fetchProductionWithMeta(
 
 const MAX_PAGE_SIZE = 100;
 const MAX_SEARCH_LENGTH = 200;
+const MAX_SEARCH_TERMS = 20;
+
+const SearchParamSchema = z
+  .preprocess((val: unknown): string[] | undefined => {
+    if (val === undefined || val === null) return undefined;
+    const raw = Array.isArray(val) ? val : [val];
+    const trimmed = raw
+      .filter((x): x is string => typeof x === "string")
+      .map((s) => s.trim())
+      .filter((s) => s.length > 0);
+    return trimmed.length === 0 ? undefined : trimmed;
+  }, z.array(z.string().max(MAX_SEARCH_LENGTH)).max(MAX_SEARCH_TERMS).optional())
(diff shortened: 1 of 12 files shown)


lucasodb:
I forgot to make it so that a search query is persistent when refreshing, just like the page index. This should only be a very small fix.

lucasodb:
> I forgot to make it so that a search query is persistent when refreshing, just like the page index. This should only be a very small fix.

Added.

backend/src/routes/production/handlers/fetch.ts
noahvsb, line 1: This file is getting pretty long, split it up. For the new file names I would suggest using `/productions/helpers/search.ts` and `/productions/helpers/pagination.ts`, but you have a full freedom here.

backend/src/routes/production/production.ts
```diff
@@ -15,7 +15,7 @@ import {
  * Registers production routes on the Fastify instance.
  *
  * @remarks
- * - `GET /api/v1/production` — fetch a list of productions.
+ * - `GET /api/v1/production` — fetch a list of productions (optional `limit`, `offset`, `search`).
```
noahvsb: You should note that these optionals should be URL parameters and not body fields.

noahvsb (changes requested):
Image

With this styling, I thought you would be able to have multiple search terms, but you don't. So either change the styling or add an option to have multiple search terms.

noahvsb (changes requested):
Requested changes look good, apart from one small thing in the URL:

Currently when having multiple search terms it looks like `?search=term1&search=term2`, but this didn't feel right. So I did some research and according to [this google search documentation](https://developers.google.com/search/docs/crawling-indexing/url-structure) we should be using comma's to list multiple values for 1 key. My previous example would then be `?search=term1,term2`.

noahvsb (approved):
Well done

Issue #308, opened 2026-04-11
[DEV] Validation errors return generic message instead of field-level details


Simon-VdB:
**Location**

All backend endpoints error handeling

____

**Description**

When doing POST to an endpoint with a Zod validation error, you only get `{"error": "Invalid request data"}`.
This should give you more detail about what actually went wrong.
Proposal: return Zod issues together with the error message (400 errors).

____

**Sources**

N/A

PR #309, opened 2026-04-11
feat: include Zod validation details in 400 error responses

Simon-VdB:
**Issue(s)**

https://github.com/SELab-2/viernulvier-1/issues/308

____

**Description**

When doing POST to an endpoint with a Zod validation error, you now get more detail about what actually went wrong.

____

**Sources**

N/A

### DIFF
--- a/backend/src/routes/helpers.ts
+++ b/backend/src/routes/helpers.ts
@@ -114,8 +114,8 @@ export const enum ParseContext {
 
 type ParseContextType = (typeof ParseContext)[keyof typeof ParseContext];
 
-function createParseError(context: ParseContextType, error: z.ZodError): HttpError {
-  if (context === ParseContext.Request) {
+function createParseError(context: ParseContextType, error?: z.ZodError): HttpError {
+  if (context === ParseContext.Request && error) {
     return new ValidationError(error.issues);
   }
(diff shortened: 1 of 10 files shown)


backend/src/routes/helpers.ts
noahvsb, line 1: I think you can just change the `parseErrors` to throw a `ValidationError` on request instead.

backend/test/routes/helpers.test.ts
noahvsb, line 1: I would make these tests expect a full match which also matches the error details.

noahvsb (changes requested):
Very helpful, thanks a lot <3

I would also like you to include these extra details in the `ApiError` class in the front-end.

backend/src/routes/helpers.ts
```diff
@@ -257,7 +259,7 @@ export function buildQuery<
       );
     } catch (err) {
       server.log.error(err);
-      throw parseErrors[ParseContext.Database];
+      throw new HttpError(HttpServerError.InternalServerError, "Internal server error");
```
noahvsb:
```suggestion
      throw createParseError(ParseContext.Database, err);
```

noahvsb (changes requested):

noahvsb (approved):
Correctly implemented requested changes

PR #310, opened 2026-04-12
filters for productions page

### LINKED ISSUE #292: [DEV] filters for productions page
**Location**

`frontend/src/utils`, `frontend/src/views/ProductionsView.vue`

____

**Description**

Create filters to search production on date, tags, etc.

____

**Sources**

lucasodb:
**Issue(s)**

Closes #292 

____

**Description**

This PR adds tag and date filters to the productions page. 

It's quite big but it's not as intimidating as it looks. It's difficult to split up something like this, and I also keep thinking of improvements while testing, so there will probably be a lot more small additions. The most important parts are of course the `ProductionsView`, and the added `ProductionsDateFilter`.  Then there are changes to the fetch logic, because of the added query parameters for tags, and for the year-range and calendar period.

I am still waiting for the scraper to be finished, so I can test this page will large amount of real API data. Currently you can only filter on 'genres', so tags of tag_type genre, because all the legacy data tags are set to be genres. I also don't like the current layout that displays the entire list of tags, and when the real data gets added there will be more than 100, so then we should definitely do this differently. I'm not sure what would be best, just a 'view more' button that expands to show all filters is also not the right approach I believe. I think it would be best to create some kind of internal mapping, that maps multiple different tags to a tag that logically includes all of them.

Also I'm not sure if the tag filters should be AND or OR.

-  (make date of calendar component always be dd/mm/yyyy instead of machine default)
-  (split up ProductionsView ? (very big file, but I think this would be very annoying to split up))
- [x] make tag filters not only be for genres

Best in seperate pr's:

-  staggered row animation
- filter results on name/date ?
- parsing and mapping of tags

____

**Sources**


Simon-VdB:
When I visited the localhost I immediately got this error:
```
4:02:29 PM [vite] Internal server error: Invalid declaration: t("nav.home")
viernulvier-frontend  |   Plugin: @tailwindcss/vite:generate:serve
viernulvier-frontend  |   File: /app/frontend/src/components/AppNavbar.vue?vue&type=style&index=0&scoped=2694e7c3&lang.css
viernulvier-frontend  |       at Se (file:///app/node_modules/tailwindcss/dist/chunk-F4544Y4M.mjs:1:3315)
viernulvier-frontend  |       at yf (file:///app/node_modules/tailwindcss/dist/chunk-F4544Y4M.mjs:38:1384)
viernulvier-frontend  |       at uu (file:///app/node_modules/@tailwindcss/node/dist/index.mjs:10:3463)
viernulvier-frontend  |       at B.generate (file:///app/node_modules/@tailwindcss/vite/dist/index.mjs:1:5242)
viernulvier-frontend  |       at TransformPluginContext.handler (file:///app/node_modules/@tailwindcss/vite/dist/index.mjs:1:2726)
viernulvier-frontend  |       at EnvironmentPluginContainer.transform (file:///app/node_modules/vite/dist/node/chunks/node.js:30048:51)
viernulvier-frontend  |       at loadAndTransform (file:///app/node_modules/vite/dist/node/chunks/node.js:24177:48)
```

I asked claude what was happening and this was his response:
"
Het probleem zit in frontend/src/components/productions/ProductionsDateFilter.vue op regel 435 en 439:
```
:global(.dark) .year-range-thumb::-webkit-slider-thumb { ... }
:global(.dark) .year-range-thumb::-moz-range-thumb { ... }
```
:global() is een Vue-specifieke pseudo-selector voor scoped styles — Vue transformeert die, maar Tailwind's v4 Vite-plugin probeert de CSS te parsen voordat Vue de transform doet, en struikelt over :global() omdat dat geen geldige CSS is.

De error-boodschap is misleidend: Tailwind wijst naar het (lege) <style scoped> blok van AppNavbar.vue en klaagt over t("nav.home"), maar dat is een parse-overflow — de parser raakt uit de bocht bij ProductionsDateFilter.vue en komt terecht in de template van een ander component.

De standaard Vue 3 manier om dit te doen is:
```
.dark :deep(.year-range-thumb::-webkit-slider-thumb) { ... }
```
of een plain .dark .year-range-thumb::-webkit-slider-thumb combinator (omdat .dark op <html> staat, buiten scope).
"
Hopefully it's clear what the problem is and how to fix it. If not let me know.

Simon-VdB:
This is some really nice work. I like how you approached this. I had a few thoughts when testing, but these are mostly details. Feel free to argue. 

**Improvements**
- Year slider with only 2026
When there is only one year of data, the slider shows “2026 – 2026” — confusing since there’s nothing to filter. Hide the slider (or disable it) when min == max, or simply display “Only 2026 available”.

- The intentions of the divider between “Filter by years” and “Filter by day / period” are a bit unclear: do you filter on both or one or the other? If "or" is the right answer, maybe you could make this more clear by having OR in the divider? (f.e. like this 
image)

- “Select dates” button doesn’t reflect active state
When filters are active and the panel is closed, the button looks the same as without filters. Add a badge/counter (“Select dates (1)”) or give it a different accent color.

- Day/period instructions
“Pick a start date then an end date (or set both fields)” feels messy. Shorter: “Pick both a start and end date”.

- Empty state
“No productions match these filters.” as plain text feels cold. Add:
	•	A small icon (e.g. empty calendar page)
	•	A suggestion: “Try broadening your date range” or a “Clear filters” button
	

Small details:
	•	The pill “16/04/2026 – 30/04/2026” uses /, while the input shows 15-04-2026. Pick one format.
	•	“Filter by years” → vague term. “Year range” is clearer.

backend/src/routes/production/helpers/pagination.ts
```diff
@@ -3,17 +3,46 @@ import { SearchParamSchema } from "./search.js";
 
 export const MAX_PAGE_SIZE = 100;
 
-/**
- * Parsed query for `GET /production`: pagination (`limit` / `offset`), optional `search` terms, and optional `old_id` filter.
- */
+/** Zod refine message for `from` / `to` order; also used in API responses and client copy mapping. */
+export const PRODUCTION_LIST_DATE_RANGE_ORDER_MESSAGE =
```
Simon-VdB: This is fragile — any typo fix on the backend silently breaks user-friendly error mapping on the client. Suggestion: export these constants from shared/ (they're literal strings, trivially shareable), or return a machine-readable error code from the backend (code: "DATE_RANGE_ORDER") and match on that instead of err.message.

backend/src/routes/production/helpers/list-where.ts
Simon-VdB, line 1: 97 lines of SQL-building logic with edge cases (parsePositiveIdList capping, dedup, escape patterns, the legacy-old_id short-circuit, empty-parts return) — covered only indirectly through handler tests. A dedicated list-where.test.ts checking each branch would make future refactors safer.

frontend/src/components/productions/ProductionsDateFilter.vue
Simon-VdB, line 1: (461 lines, non-trivial pointer-drag state) has zero tests. At minimum, test onDateInputsChange auto-swap, commitYearRange null-when-full-range logic, and mutual exclusivity with dates.

Simon-VdB (changes requested):
Thank you for the changes. 
One thing that still isn't clear to me is if the 2 parts of the date filter are AND or OR. I think it should be more obvious for the user.
Overall looks very good!

Simon-VdB (approved):
Nice work on this pass — all the review blockers from last round are gone. The shared error-code module is a cleaner pattern than what I originally suggested (machine-readable codes + message fallbacks, flowing cleanly through HttpError → JSON → ApiError → UI), and the test coverage is now solid: direct unit tests for list-where.ts and the date-filter model, plus the big integration tests for ProductionsView covering URL loading, chip removal, error-code mapping, and the expand toggles. Extracting productionsDateFilterModel.ts as pure helpers was a good call — it makes that logic testable without mounting Vue.

Verdict: ready to merge.

Three small things worth noting (none blocking):

Raw error message fallback in failListAttempt — when a 400 comes back that isn't one of the two mapped codes/messages, loadErrorDetail.value = err.message shows the raw backend string to the user. That's usually untranslated English and can leak technical wording. I'd swap it for the generic t("productionsPage.error") so the UX stays consistent and i18n-clean. One-line change.
Tag filter is AND, not OR — you flagged this question in the original description. It's currently AND (every selected tag must match), which is what list-where.test.ts locks in. That's a defensible choice — just want to make sure it's intentional rather than accidental. For a public archive, OR tends to be more discovery-friendly (users select "theatre" + "dance" expecting to see either), while AND is stricter and more prone to empty results. You can always iterate after merge based on how it feels with real data.
ProductionsView.vue size — you already flagged this. Not a blocker, just a reminder that extracting useProductionsFilterState or an ActiveFiltersBar component will get easier now that the tests are in place to catch regressions.
Good PR. 👍

PR #314, opened 2026-04-12
Feat/scraper productions

### LINKED ISSUE #213: [DEV] Scraper Productions
**Location**

Backend

____

**Description**

Scrape the viernulvier live site for past productions. Needs to have a bulk and single id variant. (Bulk for initial scrape, update to keep it up to date)

____

**Sources**

[API](https://www.viernulvier.gent/api/docs)

### LINKED ISSUE #214: [DEV] Scraper Events

### LINKED ISSUE #215: [DEV] Scraper Hall and Pricing

### LINKED ISSUE #325: [DEV] Scraper Tags
**Location**

Backend

____

**Description**

Scrape the tags for every scraped production.

____

**Sources**

msth5:
**Issue(s)**

Closes #214 
Closes #213 
Closes #215 
Closes #325 

____

**Description**

Combination of all 3 of the branches of the issues above, as they needed to be integrated anyway

____

**Sources**


lucasodb:
I will take over this pr.

Here is a list of problems/possible improvements that I've noticed:

- [x] Right now this code only really supports full database fill behaviour. We'll need to make some changes to be able to also handle the daily fetching of new events, like 'skip if exists' logic, so we don't need two unrelated implementations of the same API mapping. 

- [x] `const totalPages = meta.view.last.split("page=")[1] as unknown as number;` is wrong, because this will give a string, not a number. JS might still make this work because of type coercion, but it's not safe.

- [x] It is assumed that the `ends_at` and `doors_at` fields are always present for every event, which is not the case. 

- [x] It is assumes that `space` is always present for halls, so we shouldn't call `fetchSpaceLocation` when this isn't the case (even though this doesn't really cause a problem).

- [x] `processProduction` spreads `...production` into the POST body. It's safter to build an explicit DTO that matches `CreateProductionBodySchema`.

- [x] `login("admin", "password")` and `http://localhost:3000` are hard-coded in four places. Surely there is a better way to handle this.

- [x] It makes sense to also add the scraper for tags in this pr, blog(post) can wait though.

- [x] There are some unused imports.

I will try to fix all of these.

msth5:
Reason for lazy loading: Events the only endpoint to which you can filter a date. If u fully load productions, u need to filter these down to show on site as one doesn't wanna show future productions.
To me no point to save rows into DB that are not used and again, event is only endpoint to filter.

I was personally thinking to have a way to either extract or have as input last scrape day as client was talking about having script on recurring base, but they talked on a week bases. To have flexibilty I'd do what I said earlier.

lucasodb:
> Reason for lazy loading: Events the only endpoint to which you can filter a date. If u fully load productions, u need to filter these down to show on site as one doesn't wanna show future productions. To me no point to save rows into DB that are not used and again, event is only endpoint to filter.
> 
> I was personally thinking to have a way to either extract or have as input last scrape day as client was talking about having script on recurring base, but they talked on a week bases. To have flexibilty I'd do what I said earlier.

You're very right. I was thinking of some other way of handling it, where we would have some 'future' data in db and then filter on it, but this really would just complicate things unecessarily. Fetching on endpoints seperately is way more efficient than lazy fetching, but since you very rarely do a 'full fill', this doesn't really matter. I'll reimplement the fetching logic the way you did it.

backend/src/scraper/hall.ts
```diff
@@ -0,0 +1,387 @@
...
+
+/**
+ * `POST`s one hall parsed from Viernulvier JSON-LD into our API.
+ */
+async function createLocalHallFromViernulvierJson(
+  hall: HallJSON,
+  loginToken: string,
+  authToken: string,
+): Promise<number | null> {
+  const id = parseInt(hall["@id"].split("/").pop() as string, 10);
```
Simon-VdB: If @id is malformed and pop() returns something non-numeric, id will be NaN and the old_id: id in the POST body could cause silent failures.
In event.ts you handle this check with Number.isFinite(id) :
```
  const idSegment = event["@id"].split("/").pop();
  const id = idSegment !== undefined ? parseInt(idSegment, 10) : Number.NaN;
  if (!Number.isFinite(id)) {
    stats.events.skippedInvalidEventId++;
    console.warn(`Skipping event: could not parse legacy id from ${event["@id"]}`);
    return;
  }
```

backend/src/scraper/hall.ts
```diff
@@ -0,0 +1,387 @@
...
+
+/**
+ * Creates the hall locally when `old_id` is absent; otherwise returns the existing id.
+ */
+async function ensureHallImported(
+  hall: HallJSON,
+  loginToken: string,
+  authToken: string,
+): Promise<number | null> {
+  const oldId = parseInt(hall["@id"].split("/").pop() as string, 10);
```
Simon-VdB: same here: need some checks like Number.isFinite(...) i think

backend/src/scraper/production.ts
```diff
@@ -0,0 +1,321 @@
...
+    throw new Error(`Multiple productions found with old_id ${oldId}`);
+  }
+  return data.items[0]!.id;
+}
+
+async function createLocalProductionFromViernulvierJson(
+  production: ProductionJSON,
+  loginToken: string,
+): Promise<number | null> {
+  const id = parseInt(production["@id"].split("/").pop() as string, 10);
```
Simon-VdB: Same here: need some checks like Number.isFinite(...) i think.

backend/src/scraper/production.ts
```diff
@@ -0,0 +1,321 @@
...
+/**
+ * Creates the production locally when `old_id` is absent; otherwise returns the existing id.
+ */
+async function ensureProductionImported(
+  production: ProductionJSON,
+  loginToken: string,
+  authToken: string,
+  stats?: ScrapeRunStats,
+): Promise<number | null> {
+  const oldId = parseInt(production["@id"].split("/").pop() as string, 10);
```
Simon-VdB: Same here: need some checks like Number.isFinite(...) i think.

backend/eslint.config.js
```diff
@@ -128,6 +128,15 @@ export default defineConfig([
     },
   },
 
+  // Archive scraper: external API JSON → trusted numeric keys; stats path from env / fixed default.
+  {
+    files: ["src/scraper/**/*.ts"],
+    rules: {
+      "security/detect-non-literal-fs-filename": "off",
+      "security/detect-object-injection": "off",
```
Simon-VdB, line 136: Blanket-disabling detect-object-injection for the entire scraper directory is broad — could you narrow it to the specific files that actually use dynamic keys (scrape-stats.ts, language-map.ts)?

backend/src/scraper/event.ts
```diff
@@ -0,0 +1,384 @@
...
+  const prices = priceUrls.map((priceUrl) => parseInt(priceUrl.split("/").pop() as string, 10));
+  if (prices.length > 0) {
+    await scrapeEventPricesForEvent(prices, eventId, authToken, loginToken);
+  }
+}
+
+/**
+ * Walks every page of the (bounded) Viernulvier events list and runs {@link ensureEventImported} per member.
+ */
+export async function scrapeAllEvents(
```
Simon-VdB, line 360: very clean

backend/src/routes/event/handlers/helper.ts
noahvsb, line 30: This is probably not necessary because we're using `coerce` in the typing.

backend/src/routes/hall/handlers/create.ts
noahvsb, line 1: RIP comments 🙏

backend/src/routes/hall/handlers/replace.ts
noahvsb, line 1: Comments gone here as well.

backend/src/routes/production/handlers/body-schema.ts
```diff
@@ -25,6 +25,10 @@ export const CreateProductionBodySchema = ProductionSchema.pick({
   artist: true,
   tagline: true,
   teaser: true,
+}).partial({
+  artist: true,
+  tagline: true,
+  teaser: true,
```
noahvsb: This part feels odd to me, can you explain why these are both picked as normal and as partial?

backend/src/scraper/auth.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+/**
+ * Password paired with {@link scraperAdminUsername} for `POST /api/v1/auth/login`.
+ * Set `SCRAPER_ADMIN_PASSWORD` in any environment where the default is not acceptable.
+ */
+export function scraperAdminPassword(): string {
+  const v = process.env["SCRAPER_ADMIN_PASSWORD"]?.trim();
+  return v && v.length > 0 ? v : "password";
+}
```
noahvsb: I would merge the username and password function in one. Looking at the docstring of the username function it looks like this was intended at first.

Simon-VdB (changes requested):
This is a solid, well-thought-out PR. The scraper architecture is clean, the API changes are backward-compatible and necessary, ... The main items I'd want addressed before merging:

Bug: Add NaN guards to parseInt in hall/production scraper code (match the pattern already used in event.ts)
Hygiene: Narrow the ESLint security rule suppression if possible
When I tested the previous-brussels-day i got this error:
```
simonvandenberghe@MacBook-Pro-van-Simon backend % SCRAPE_EVENTS_WINDOW=previous-brussels-day pnpm run scrape
 WARN  Unsupported engine: wanted: {"node":">=24"} (current: {"node":"v20.19.1","pnpm":"10.30.3"})

> backend@1.0.0 scrape /Users/simonvandenberghe/Ugent/viernulvier-1/backend
> tsx scripts/scrape.ts

[dotenv@17.3.1] injecting env (7) from ../.env -- tip: ⚙️  write to custom object with { processEnv: myObject }
Scraping events (lazy halls/productions)… window: previous-brussels-day; after=2026-04-15T22:00:00.000Z before=2026-04-16T22:00:00.000Z
TypeError: Cannot read properties of undefined (reading 'length')
    at totalPagesFromHydraView (/Users/simonvandenberghe/Ugent/viernulvier-1/backend/src/scraper/hydra-view.ts:48:11)
    at scrapeAllEvents (/Users/simonvandenberghe/Ugent/viernulvier-1/backend/src/scraper/event.ts:368:22)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async main (/Users/simonvandenberghe/Ugent/viernulvier-1/backend/scripts/scrape.ts:64:17)
 ELIFECYCLE  Command failed with exit code 1.
```
The crash is on line 48: iri.length where iri is undefined. That means view.last is undefined and view.first is also undefined.

This happens when the Viernulvier API returns zero results for the time window. When there are no events for "yesterday" (April 16), the API likely returns a response with no view object at all, or a view without first/last. The code in fetchEventsListMeta blindly casts the response to EventListMeta and assumes view.first exists.

Look at the EventListMeta interface in event.ts:43:
```
interface EventListMeta {
  view: {
    first: string;
    last?: string;
  };
}
```
It treats first as always present, but with an empty result set the API probably returns either no view at all, or view exists but has no first/last.

Overall great work! 🫡

Simon-VdB (commented):
Nice work, the error is gone now, so thank you. All requested changes look good.

backend/src/scraper/language-map.ts
```diff
@@ -0,0 +1,46 @@
...
+    .replace(/&apos;/gi, "'")
+    .replace(/&lt;/gi, "<")
+    .replace(/&gt;/gi, ">")
+    .replace(/&amp;/gi, "&");
+  return s.replace(/\s+/g, " ").trim();
+}
+
+/**
+ * Keeps only allowed language keys with non-empty trimmed strings.
+ * Returns `null` when the result would be `{}` (our Zod `languageMap` rejects empty objects).
```
noahvsb, line 32: If our zod rejects empty objects and that's wrong, then maybe we should change the zod instead of the input?

backend/src/scraper/hall.ts
```diff
@@ -0,0 +1,397 @@
...
+  return data;
+}
+
+/**
+ * Maps space resource URL (trimmed) → resolved address or `null` when unknown (avoids duplicate venue fetches).
+ */
+const spaceAddressCache = new Map<string, string | null>();
+
+/** Hydra / JSON-LD link: plain IRI string or `{ "@id": "..." }`. */
+export function hydraIriString(ref: unknown): string | null {
```
noahvsb: Move this to `hydra-view.ts`

backend/src/scraper/scrape-stats.ts
```diff
@@ -0,0 +1,148 @@
...
+  const v = process.env["SCRAPE_VERBOSE"]?.trim().toLowerCase();
+  return v === "1" || v === "true" || v === "yes";
+}
+
+/** Repo root (`viernulvier/`), from `backend/src/scraper/*.ts`. */
+const repoRoot = path.resolve(path.dirname(fileURLToPath(import.meta.url)), "..", "..", "..");
+
+/**
+ * Where to write {@link formatRunReport} output. Override with `SCRAPE_STATS_FILE` (absolute or relative to cwd).
+ * Default file: `backend/scripts/last-scrape-stats.txt` (next to `scripts/scrape.ts`).
```
noahvsb: I would default to `backend/out/last-scrape-stats.log`

backend/vitest.config.ts
```diff
@@ -24,7 +24,7 @@ export default defineConfig({
     coverage: {
       provider: "v8",
       include: ["src/**/*.{ts,tsx}"],
-      exclude: ["src/index.ts"],
+      exclude: ["src/index.ts", "src/scraper/**"],
```
noahvsb: I would like you to include this just to see it in the coverage overview. You can set the thresholds for this to 0 though for now. It's against my principles to approve this PR without proper tests, but we're getting behind on schedule.

frontend/src/views/ProductionsView.vue
noahvsb, line 1: This file is insanely big, can you split it up in components like the home page in #292?

.gitignore
noahvsb, line 1: When you do the requested change on `backend/src/scraper/scrape-stats.ts`, you should add `backend/out` here.

noahvsb (changes requested):
Very nicely split up code, well documented.

I didn't test it locally, since Simon already did that.

Apart from the changes on the files, I would also like some documentation in a markdown on how to use the scraper.

And also for the local dev environment, please make a pnpm script that fetches only the last month or so. You don't need all of the productions in the local dev environment. And a bash script that does docker exec like the other bash scripts.

backend/src/routes/hall/handlers/create.ts
lucasodb: Yes, the comments shouldn't be there, and are also not there in all other handler files, so I removed them for consistency.

backend/src/routes/production/handlers/body-schema.ts
lucasodb: You're right, I should've removed them from the pick. Actually, I think I can just do pick for title and then add the other fields to the extend, that would be even cleaner.

frontend/src/views/ProductionsView.vue
lucasodb: That's definitely not something for here, I mentiond in #310 that I was still planning to do that there.

backend/src/scraper/language-map.ts
lucasodb: If we allowed an empty object in the zod, we'd have to treat {} like null everywhere else and account for this, so I think it's better handled this way.

backend/src/routes/event/handlers/helper.ts
lucasodb: Hmm, there is indeed some double work here, yes. But I think with only zod, you get an error on an invalid date, instead of keeping the old value, so this is still meaningful to better handle invalid dates. Maybe just make a tiny seperate pr for this if you know the best way to handle this.

noahvsb: I'll make an issue on this.

noahvsb (approved):
Implementation of requested changes look good to me 👍

Simon-VdB (approved):

PR #326, opened 2026-04-17
fixed event prices path in services

trherber:
**Issue(s)**

____

**Description**

The frontend service was using the deprecated /event_price path instead of the new /event/price path. 

____

**Sources**


noahvsb (approved):
Looks good to me, nice catch

Issue #330, opened 2026-04-18
[DEV] CMS tabs


Simon-VdB:
**Description**

Make 3 tabs for productions, tags and admins in the CMS. Then the CMSView has 3 components inside for each of those tabs.

____

**Sources**

N/A

PR #331, opened 2026-04-18
added tabs

Simon-VdB:
**Issue(s)**

https://github.com/SELab-2/viernulvier-1/issues/330

____

**Description**

Added the tabs in the CMSView and made 3 placeholder components.

____

**Sources**

N/A


frontend/src/views/admin/CMSView.vue
```diff
@@ -14,999 +13,56 @@
...
+  { id: "admins", labelKey: "cms.tabs.admins" },
+];
 
-onMounted(() => {
-  window.addEventListener("keydown", onWindowKeyDown, true);
-  resetCreateLinkedEventForm();
-  void loadCmsData();
-});
+const activeTab = ref<CmsTabId>("productions");
+const productionsTabRef = ref<InstanceType<typeof CmsProductionsTab> | null>(null);
```
noahvsb: Linter warning.

```
  65:27  warning  Replace 'ref' with 'useTemplateRef'  vue/prefer-use-template-ref
```

frontend/src/components/admin/cms/CmsProductionsTab.vue
```diff
@@ -0,0 +1,980 @@
...
+  if (!eventsPanelError.value) {
+    createEventModalOpen.value = false;
+  }
+}
+
+function getInitialDark(): boolean {
+  const stored = localStorage.getItem("viernulvier-dark");
+  if (stored !== null) return stored === "true";
+  return window.matchMedia("(prefers-color-scheme: dark)").matches;
+}
```
noahvsb: `useDarkMode.ts` has this function, why is it explicitly written here? Just import it.

noahvsb (changes requested):
Looks good, code is nice! But just a few things:

Like with the blogpost detail page, the footer is in the middle of the screen, it should be at the bottom. This is out of scope for this PR, because I feel like there will be more locations with the issue, but can you make an issue on this?

For the file locations I would split `components/admin/cms` in 3 subfolders: `tabs`, `admins` and `productions`. The components that can be used for all can stay in their current location, the rest needs to move in their respective subfolder.

noahvsb (approved):
Looks good now 🫡

PR #343, opened 2026-04-19
handle overlap of api/legacy data

lucasodb:
**Issue(s)**

____

**Description**

This pr is to improve the way legacy production imports happen, when the db is already populated with scraped data from the official api. For each row in the csv data files we now first check if there is a production present in the database that has the same title and artist, and if so, this production skipped and not inserted.

There is still a problem with this approach though. VIERNULVIER has a few recurring events, mostly raves etc., and some of them have hundreds of productions in the api/legacy data, like 'New Wave Classix' and 'Hindu Nights'. The problem is that the production title for these is always the same, and the artist too (empty most of the time), so by just matching on these two fields, a ton of productions are not added.

So, you would also need to  match on event dates, but that gets a little complicated, so I'm not sure what the best way is to handle that yet.

____

**Sources**

### DIFF
--- a/backend/scripts/import-events-legacy.ts
+++ b/backend/scripts/import-events-legacy.ts
@@ -13,15 +13,26 @@ const DEFAULT_FILE = resolveDefaultLegacyImportFile(import.meta.url, [
   "events.csv",
 ]);
 
+const DEFAULT_PRODUCTIONS_FILE = resolveDefaultLegacyImportFile(import.meta.url, [
+  "data",
+  "imports",
+  "productions.csv",
+]);
+
 async function main() {
   const args = parseLegacyImportCli({
     defaultFile: DEFAULT_FILE,
     scriptForUsage: "import:events",
     description: "Import legacy events CSV into Postgres.",
+    eventsProductionsDefault: DEFAULT_PRODUCTIONS_FILE,
   });
 
   assertDatabaseUrl();
   assertCsvFileExists(args.filePath);
+  if (args.productionsFilePath === undefined) {
+    throw new Error("Internal error: missing productionsFilePath (events CLI must set eventsProductionsDefault).");
+  }
+  assertCsvFileExists(args.productionsFilePath);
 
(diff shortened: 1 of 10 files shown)


lucasodb:
Improved way of handling overlapping data:

Production import no longer treats “same title + artist” as a single match. We only map a legacy row to an existing production when exactly one DB row matches, because then we can be fairly sure that this is actually the same production. If there are zero or several matches, we insert a new production and record that in legacy_production_import_map (created_new_production), so recurring editions don’t get merged by mistake. 

Event import now loads the productions CSV (same title/artist metadata) and groups event rows by legacy production id. For each group we compare calendar days against event.starts_at on matching API productions. If any legacy date already exists on the scraped/API side for that production, we treat the whole legacy production as redundant: we skip all its events, and if we had inserted an orphan production for an ambiguous title/artist match, we delete that row so the DB stays clean.

Here is what my output looked like of running the two scripts:

**pnpm run import:productions**:

  Rows read: 6129
  Imported productions: 4660
  Created genre tags: 0
  Linked production_tag rows: 4862
  Skipped (already imported): 0
  Skipped (missing legacy ID): 244
  Skipped (duplicate legacy ID in file): 0
  Skipped (missing title): 213
  Skipped (Zod validation failed): 0
  Legacy IDs mapped to existing productions (exactly one title+artist match, no new row): 1012
  New production rows (zero or multiple title+artist matches): 4660
  Failed rows: 0
  
 **pnpm run import:events**:
 
  Rows read: 10598
  Imported events: 9951
  Created halls: 45
  Skipped (full duplicate production vs API calendar day): 269
  Deleted orphan productions after duplicate detection: 11
  Skipped (missing starttime): 1
  Skipped (missing hall): 94
  Skipped (missing production legacy id): 0
  Skipped (unknown production mapping): 227
  Skipped (duplicate in file): 56
  Skipped (already imported): 0
  Skipped (Zod validation failed): 0
  Failed rows: 0

msth5 (approved):
Sometimes a bit hard to get the head around, but seems a decent addition that implements what was set as goal

Issue #336, opened 2026-04-19
[DEV] admins tab in CMS


Simon-VdB:
**Location**

`/frontend/src/components/admin/cms/tabs/CmsAdminsTab.vue`

____

**Description**

Implement the ag grid for the admins tab in the admin cms.

____

**Sources**

`/frontend/src/components/admin/cms/tabs/CmsProductionsTab.vue`

PR #344, opened 2026-04-19
implemented admin tab

noahvsb:
**Issue(s)**

#336 

____

**Description**

Implements the admin tab.

I also implemented a URL parameter for the tabs to make it more stateless: when you're on the admins tab, you're on `/en/admin/cms`, so when you refresh you get back to the default productions tab. So now when you're on the admins tab, you're on `/en/admin/cms?tab=admin` and when you refresh you stay there. Idem for tags.

____

**Sources**


noahvsb:
Why are the checks not executing.

frontend/src/composables/useCmsAdminGrid.ts
```diff
@@ -0,0 +1,279 @@
...
+    sortable: true,
+    filter: true,
+    floatingFilter: true,
+    resizable: true,
+    minWidth: 120,
+  };
+
+  const gridColumnOptions = computed(() => [
+    { colId: "username", label: options.t("cms.columns.admin.username") },
+    { colId: "profilePicture", label: options.t("cms.columns.admin.profilePicture") },
```
willemDC-Info: In the grid there is a column for the admin to have a profile picture however there is no way to upload a picture for the admin in the create tab.

Either remove the column or give the admin to upload an image.

frontend/src/services/cms/types.ts
```diff
@@ -37,6 +37,14 @@ export interface CmsProductionGridRow {
   events: number[];
 }
 
+export interface CmsAdminGridRow {
+  id: number;
+  source: Admin;
+  username: string;
+  profilePicture: string | null;
```
willemDC-Info: Same as before. Remove this or give the admin the possibiliy to upload an image.

frontend/src/services/cms/types.ts
```diff
@@ -65,6 +73,12 @@ export interface CreateFormState {
   video_2: Record<SupportedLang, string>;
 }
 
+export interface CreateAdminFormState {
+  username: string;
+  password: string;
+  super: boolean;
+}
```
willemDC-Info, line 98: idem

willemDC-Info (changes requested):

frontend/src/composables/useCmsAdminGrid.ts
noahvsb: The back-end doesn't have support for adding a profile picture when creating or editing an admin, so that's why it's not added to the create modal. I'll just comment the code out here and in the other locations so you can easily uncomment when there is back-end support.

willemDC-Info (dismissed):
The code looks good to me and runs well on the frontend.
I will merge it to staging after the merge conflict have been resolved

willemDC-Info (approved):
Code looks all good and works fine. Good job!

PR #349, opened 2026-04-21
Feat/productions page improvements

lucasodb:
**Issue(s)**

____

**Description**

Just some small improvements to the productions page

- avoid layout issues by waiting for events on the current page to be finished fetching, before production cards are shown
- all genres are shown first on the production cards
- tag filters are now OR instead of AND

____

**Sources**

### DIFF
--- a/frontend/src/views/ProductionsView.vue
+++ b/frontend/src/views/ProductionsView.vue
@@ -194,8 +194,8 @@
           <div
             v-if="
               filterBannerTagIds.length > 0 ||
-              filterBannerYearRange !== null ||
-              (filterBannerDateFrom && filterBannerDateTo)
+                filterBannerYearRange !== null ||
+                (filterBannerDateFrom && filterBannerDateTo)
             "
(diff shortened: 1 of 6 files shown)


frontend/src/views/ProductionsView.vue
Simon-VdB, line 1: Some minor linting warnings here.

Simon-VdB: I see these where already there, but no problem, you can just fix them now if that’s alright with you?

Simon-VdB (changes requested):
Very nice work, I love that you’re refactoring this to make it cleaner both logically on the frontend, but also the code. It looks all good to me, just a heads up that I didn’t build or tested the things, because I’m reviewing this on my ipad during class, but I think if it works on you’re computer these changes don’t seem to be able to break it on another computer.

Simon-VdB (approved):
Awsome, thank you.

PR #350, opened 2026-04-21
add staggered entrance for production cards

lucasodb:
**Issue(s)**

____

**Description**

Added a staggered entrance for the production cards. This makes the page feel a bit more alive. Feel free to tell me if you don't like this. Also, when a filter is added, and productions currently on the page still match the search, these will just stay static, and won't re-stagger with the animation, while other productions on the page will. This doesn't really bother me, but there might be a better way to handle this.

____

**Sources**

### DIFF
--- a/frontend/src/views/ProductionsView.vue
+++ b/frontend/src/views/ProductionsView.vue
@@ -194,8 +194,8 @@
           <div
             v-if="
               filterBannerTagIds.length > 0 ||
-              filterBannerYearRange !== null ||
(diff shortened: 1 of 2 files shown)


frontend/src/views/ProductionsView.vue
noahvsb, line 1: Fix warnings.

noahvsb (changes requested):
Very satisfying, well done!

There's just some linter warnings I'd like fixed.

noahvsb (approved):

PR #356, opened 2026-04-22
feat/media scraper

### LINKED ISSUE #217: [DEV] Media Scraper

msth5:
**Issue(s)**

Closes #217 
Closes #358 

____

**Description**

Loads in images and crops from VierNulVier main site into the archive site

edit: This PR also contains the changes in #358 therefor merging this PR makes that PR antiquated

____

**Sources**

### DIFF
--- a/backend/scripts/scrape.ts
+++ b/backend/scripts/scrape.ts
@@ -7,7 +7,7 @@
  * - `VIERNULVIER_API_ORIGIN` — archive API origin (default `https://www.viernulvier.gent`).
  * - `VIERNULVIER_LOCAL_API_URL` — own API base (default `http://localhost:3000`).
  * - `SCRAPER_ADMIN_USERNAME` / `SCRAPER_ADMIN_PASSWORD` — JWT login for protected `POST`s (defaults `admin` / `password`).
- * - `SCRAPE_EVENTS_WINDOW` — `historical` | `previous-brussels-day`.
+ * - `SCRAPE_EVENTS_WINDOW` — `historical` | `previous-brussels-day` | `past-seven-days` | `past-thirty-days`.
  *
  * This entrypoint is events-only: halls and productions are imported lazily while processing each event.
  *
@@ -16,6 +16,8 @@
 import "@/scraper/load-repo-env.js";
 import {
   previousBrusselsDayBounds,
+  pastSevenDaysBounds,
+  pastThirtyDaysBounds,
   scrapeAllEvents,
   type ViernulvierEventStartBounds,
 } from "@/scraper/event.js";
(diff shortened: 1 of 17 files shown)


lucasodb:
Looks solid already, but I have some questions

- I think you still need to remove some dead code, that uses the old JSON path? `scrapeCrops`, `ensureCropImported` and `createLocalCropFromViernulvierJson` (and `CreateCropBody)`. `createCropsForImage` is the correct approach, so the old approach can be removed I think.

- `downloadFile` doesn't send the X-AUTH-TOKEN, doesn't it need to?

- In `fetchLocalImageIdByOldId`  you use the same idea as implemented for production, event and hall, but for media this route doesn't exist yet, there is no GET /api/v1/image without :id that accepts ?old_id=…, so this can't work. The same for `fetchLocalCropIdByOldId`.

msth5:
> Looks solid already, but I have some questions
> 
> * I think you still need to remove some dead code, that uses the old JSON path? `scrapeCrops`, `ensureCropImported` and `createLocalCropFromViernulvierJson` (and `CreateCropBody)`. `createCropsForImage` is the correct approach, so the old approach can be removed I think.

Has been fixed, was indeed either some default code from copying other scraper or old code from experimenting with the endpoints.

> * `downloadFile` doesn't send the X-AUTH-TOKEN, doesn't it need to?

These files are publicly accessible as they're needed within the html bodies. As far as I know that's pretty standard practice, therefor no authentication is necessary

lucasodb:
> > Looks solid already, but I have some questions
> > 
> > * I think you still need to remove some dead code, that uses the old JSON path? `scrapeCrops`, `ensureCropImported` and `createLocalCropFromViernulvierJson` (and `CreateCropBody)`. `createCropsForImage` is the correct approach, so the old approach can be removed I think.
> 
> Has been fixed, was indeed either some default code from copying other scraper or old code from experimenting with the endpoints.
> 
> > * `downloadFile` doesn't send the X-AUTH-TOKEN, doesn't it need to?
> 
> These files are publicly accessible as they're needed within the html bodies. As far as I know that's pretty standard practice, therefor no authentication is necessary

You're correct, the token is indeed not necessary. The are still some problems though, like pagination not being added, and the oldId not being wired correctly. Since I made a branch from this one on thursday, with all these fixes, because I quickly needed to run the script to import some example data to display in the presentation for milestone 2, you don't have to do these fixes again yourself here and we can just merge my fixes branch into this one. I had no problem with running the version with the fixes, so I think it should be all good. I just made the pr for this merge (#366).

lucasodb:
Only the crops  "FE3_header",  "FE3_home_featuredWide",  "FE3_boxed",  "nb_header" and  "cms" get fetched now.

lucasodb (approved):

PR #358, opened 2026-04-22
2 extra modes for date bounds on the scraper

msth5:
**Issue(s)**

none

____

**Description**

Two extra modes to have a local database of verious sizes, to have it loaded faster


backend/src/scraper/zoned-day.ts
noahvsb, line 35: I wouldn't call this brussels day, I know it's not part of this PR, but looking at it, it looks like UTC, which is not the brussels timezone. I wouldn't make it the actual brussels timezone, since that changes (summer- and wintertime). So I think it's best to just rename this:

```suggestion
export function previousDayBounds(): ViernulvierEventStartBounds {
```

And then add to the JSDoc that it's UTC.

noahvsb (changes requested):
Your changes look good, but I have a remark about something that was already on staging (see below). It's a small change, so that's why I thought it's best to include in this PR instead of making a new PR for this.

Issue #351, opened 2026-04-21
[DEV] frontend media service


noahvsb:
**Location**

`frontend/src/services/media.ts`

____

**Description**

Create frontend helpers for the media endpoint.
____

**Sources**

See other frontend services.

PR #365, opened 2026-04-25
Feat/media integration

### LINKED ISSUE #382: [DEV] Frontend media integration
**Location**

Productions page, productions detail page

____

**Description**

Add thumbnails to the production cards, and images to the detail page.

____

**Sources**

lucasodb:
**Issue(s)**

Closes #382, #351

____

**Description**

Add thumbnails to the production cards on the production page, and banner/carousel to the detail page.

- [x] Don't show arrows for 2 or less images.
- [x] Make images in carousel clickable to show enhanced version.
- [x] Use correct crop types.
- [x] Crops of different native resolutions look a bit weird in the carousel.

____

**Sources**


frontend/src/views/ProductionsView.vue
```diff
@@ -711,10 +714,41 @@ const hasActiveListFilters = computed(() => {
...
+  return m.get(productionId) ?? null;
+}
+
+async function loadThumbnailsForProductionIds(ids: number[]): Promise<void> {
+  if (ids.length === 0) {
+    thumbnailUrlByProductionId.value = new Map();
+    return;
+  }
+  const next = new Map<number, string | null>();
+  await Promise.all(
```
Simon-VdB, line 742: every filter/search/page interaction triggers a new fan-out of N requests. If the user clicks or types quickly, an older Promise.all may resolve after a newer one and overwrite the map with IDs that are no longer on the current page. Result: suddenly missing thumbnails (the old IDs don’t match the currently visible productions).

This is a race condition.

fix: use a request-generation counter (increment on start, check before assignment) or an AbortController. Both are standard patterns.

frontend/src/views/ProductionsView.vue
```diff
@@ -711,10 +714,41 @@ const hasActiveListFilters = computed(() => {
...
 
+function thumbnailFor(productionId: number): string | null {
+  const m = thumbnailUrlByProductionId.value;
+  if (!m.has(productionId)) {
+    return null;
+  }
+  return m.get(productionId) ?? null;
+}
+
+async function loadThumbnailsForProductionIds(ids: number[]): Promise<void> {
```
Simon-VdB, line 735: For each production on the page, a separate GET /production/:id/image request is made. With PAGE_SIZE = 20, that’s 20 parallel requests on top of the main production fetch — repeated on every filter change.
=> Poor scalability and unnecessary loads

I think a fix would be a batch endpoint (/production/images?ids=…) but that's out of scope for this PR.
If you agree we can make a follow up issue for that.

frontend/src/views/ProductionDetailView.vue
```diff
@@ -51,28 +52,41 @@ import BlogSection from "@/components/production/BlogSection.vue";
...
 const loading = ref(true);
 const error = ref<string | null>(null);
 const notFound = ref(false);
 
 onMounted(async () => {
   try {
-    production.value = await getProduction(id);
+    const [fetched, images] = await Promise.all([
+      getProduction(id),
+      getImagesForProduction(id).catch(() => [] as ImageWithCrops[]),
```
Simon-VdB: Silent api error swallowing.
no console.warn, no toast, no telemetry. A 500 error is handled exactly the same as “this production has no images”.

How to fix: at minimum log a console.warn with sufficient context (production ID + status). Optionally distinguish between 404 (expected) and 5xx (should be logged).

frontend/src/components/production/ProductionImageCarousel.vue
```diff
@@ -0,0 +1,618 @@
...
+          :aria-label="
+            t('production.gallery.goToScreen', { n: i + 1, total: screenCount })
+          "
+          :aria-current="i === activeScreenIndex ? 'true' : undefined"
+          @click="goToScreen(i)"
+        />
+      </div>
+    </div>
+
+    <Teleport to="body">
```
Simon-VdB, line 93: The lightbox is marked as role="dialog" with aria-modal="true", but that contract is not fully upheld: there is no focus trap (Tab navigation continues to elements behind the backdrop), and focus is not restored to the carousel button that opened the lightbox after closing it. This can be confusing for screen reader and keyboard users. Could we include proper focus management in this PR, or plan it as an immediate follow-up?

frontend/src/components/production/ProductionImageCarousel.vue
```diff
@@ -0,0 +1,618 @@
...
+      if (scrollerRef.value) {
+        scrollerRef.value.scrollTo({ left: 0, behavior: "auto" });
+      }
+      scheduleLayoutMetrics();
+    });
+  },
+  { deep: true },
+);
+
+watch(lightboxIndex, (idx) => {
```
Simon-VdB: On close, it blindly resets to "". If something else (route transition, mobile menu) also applies a scroll lock via inline style, that value gets wiped.
fix: store and restore the previous inline value, or toggle a class on <body> instead.

frontend/src/components/production/HeroSection.vue
```diff
@@ -1,16 +1,24 @@
 <template>
-  <section class="relative flex min-h-[70vh] w-full items-end overflow-hidden">
-    <div class="absolute inset-0 z-0 bg-surface-inv">
+  <!-- Min height matches ~1920×600 banners; flex + justify-end grows the section when overlay text needs more room -->
+  <section
+    class="relative flex w-full min-h-[max(280px,calc(100vw*600/1920))] flex-col justify-end overflow-hidden"
+  >
+    <div class="absolute inset-0 z-0 bg-black">
```
Simon-VdB: The rest of the file uses design tokens (text-ink-on-inv, bg-surface-inv-raised, …). Here it switches to bg-black where previously bg-surface-inv was used.
Is this intentional? If so, a comment would be nice, and if not, design token usage would be cleaner.

Simon-VdB (changes requested):
Nice work, it looks very good and works. I have left some questions/requested changes.

frontend/src/views/ProductionsView.vue
lucasodb: Fair, I made an issue for this #390.

Simon-VdB (approved):
Well done, thank you very much.

I approve now :)

PR #367, opened 2026-04-25
CMS refactor

### LINKED ISSUE #342: [DEV] Refactor CMS tabs to extract reusable code
**Location**

`/frontend/src/components/admin/cms/...`

____

**Description**

Reduce duplicate code across cms tabs (productions, tags, admins).

____

**Sources**

N/A

Simon-VdB:
**Issue(s)**

https://github.com/SELab-2/viernulvier-1/issues/342

____

**Description**

Reduce duplicate code across the three CMS tabs (productions, tags, admins) by extracting their shared structure into reusable building blocks.

The three tabs were structurally near-identical at two layers — grid plumbing in their composables and the surrounding template skeleton in their .vue files — but each carried its own copy. This PR consolidates both layers without changing any user-facing behavior.

1. useCmsGridBase<TRow> composable
Extracted ~180 lines of identical grid plumbing that lived in each of useCmsProductionGrid, useCmsTagGrid, and useCmsAdminGrid:

localStorage state persist/restore (loadPersistedGridState, restoreGridState, persistGridState)
Column visibility tracking and sync (syncColumnVisibilityFromGrid, setGridColumnVisibility)
Theme variables (light/dark mode color tokens)
Quick filter, fit-columns, auto-size, reset-filters, reset-state
CSV export with optional per-tab cell processor and excluded columns
onGridReady and onSelectionChanged lifecycle handlers
The three specialized composables now only declare entity-specific column definitions, default column config, and grid-column option labels — they spread the base composable's return into theirs.

A fitOnReadyAfterRestore flag preserves a subtle pre-existing behavior difference: tag/admin grids always call sizeColumnsToFit on grid-ready, while production only fits when no persisted state was restored.

2. CmsTabShell.vue wrapper component
Extracted the surrounding template skeleton that each tab repeated verbatim:

cms-tab-content wrapper
Header row with row count + action-buttons slot
CmsGridControls block (with all event wiring proxied)
Load-error banner
Optional status-banner slot (used by productions for save-success)
Grid slot (each tab supplies its own AgGridVue configuration)
CmsColumnChooser block (with event wiring proxied)
Save-error banner
Empty-state placeholder
Modals slot (drawers, side-panels, create modals)
Each tab now declares only what's actually unique: i18n keys for count/empty messages, header buttons, the grid configuration itself, and entity-specific modals/drawers.

____

**Sources**

N/A


frontend/src/components/admin/cms/productions/CmsProductionsTab.vue
noahvsb, line 1: Indentation warnings, do `pnpm run lint:fix`.

noahvsb (changes requested):
Nice refactoring, 300 lines gone.

But you should also refactor the tests, you should be able to remove a ton of lines there as well by moving the logic from the specific test files into new test files for the shell and base.

Most of the tests regarding the productions tab aren't consistent in paths with the src folder. I don't think I need to give any examples, it will be pretty clear. So please move these in the correct folders.

I would also move the tab components into their respective folders. It's confusing for example that `CmsTagsTab.vue` isn't under `cms/tags`. It also doesn't help that `tags` and `tabs` have a 1 letter difference.

noahvsb (changes requested):
Well done, but I've found two more inconsistencies in the tests:

There's a `cms-helpers.test.ts` in the `test/views/admin` folder, it doesn't belong there. It contains tests from a few files `src/services/cms/...`, so best to also split up the logic per file it has tests for.

The `CmsView.test.ts` in that same folder doesn't follow the naming convention.

noahvsb (approved):
Requested changes correctly implement, nice work

Issue #276, opened 2026-04-05
[DEV] admin home page: add a change password function


noahvsb:
**Location**

`backend/src/routes/auth/edit.ts` - add a PATCH on `/api/v1/auth/me` that expects a password field only
`frontend/src/services/auth.ts` -  add a helper function
`frontend/src/views/admin/AdminView.vue` - add a button that redirects to a new view where you edit your password
`frontend/src/views/admin/ChangePasswordView.vue`

____

**Description**

Currently super admins are allowed to edit passwords, but we should also allow non-super admins to edit their own password. This way a super admin can create an account for a stagaire for example using a random password; afterwards the stagaire can edit the password and choose their own.
____

**Sources**

PR #368, opened 2026-04-25
change own password

noahvsb:
**Issue(s)**

#276 
____

**Description**

Added backend support to change your own password (previously only super admins could do this). I felt like there was no return value necessary, so after checking [the specs](https://datatracker.ietf.org/doc/html/rfc5789#section-2.1), I went with a 204 No Content reply. Also when checking [these specs](https://datatracker.ietf.org/doc/html/rfc9110#name-www-authenticate), I encountered that a `WWW-Authenticate` header must be included on a 401 Unauthenticated reply, so I did that as well.

Then I integrated it in the frontend as a quick action with a modal. I felt like it didn't need a whole new page.

I also saw that in the SQL integration tests, the media tests were in the describe of the tags, so that's where most of my insertions and deletions went.

**TODO**
- [x] write tests for frontend

____

**Sources**

https://datatracker.ietf.org/doc/html/rfc5789#section-2.1
https://datatracker.ietf.org/doc/html/rfc9110#name-www-authenticate


frontend/src/components/admin/ChangePasswordModal.vue
```diff
@@ -0,0 +1,95 @@
...
+  loading.value = true;
+
+  try {
+    await updateOwnPassword(password.value);
+    close();
+  } catch (err) {
+    if (err instanceof ApiError) {
+      error.value = `${t('admin.changePassword.failedToUpdate')}: ${err.message}`;
+    }
+    error.value = t('admin.changePassword.failedToUpdate') + ".";
```
Simon-VdB: Shouldn't this be an else clause? Because now this overwrites the error.value every time...


```suggestion
    else {
        error.value = t('admin.changePassword.failedToUpdate') + ".";
    }
```

frontend/src/components/admin/ChangePasswordModal.vue
```diff
@@ -0,0 +1,95 @@
...
+          v-model="confirm"
+          type="password"
+          :placeholder="t('admin.changePassword.confirmPassword')"
+          class="input"
+        />
+
+        <p v-if="error" class="error">{{ error }}</p>
+
+        <div class="actions">
+          <button id="close-btn" type="button" @click="close" class="btn secondary">{{ t('general.cancel') }}</button>
```
Simon-VdB: fix eslint warning

frontend/src/components/admin/ChangePasswordModal.vue
```diff
@@ -0,0 +1,95 @@
...
+          <button id="close-btn" type="submit" class="btn primary" :disabled="loading">
```
Simon-VdB: Duplicate button id... "close-btn" used again here. This could be something like "submit-btn"

frontend/src/i18n/locales/nl.ts
```diff
@@ -1,4 +1,8 @@
 export default {
+  general: {
+    cancel: "Annuleer",
+    saving: "Toevoegen...",
```
Simon-VdB: Misschien beter: "Opslaan..."

```suggestion
    saving: "Opslaan...",
```

frontend/src/views/admin/AdminView.vue
```diff
@@ -37,27 +37,47 @@
               <line x1="5" y1="12" x2="19" y2="12" /><polyline points="12 5 19 12 12 19" />
             </svg>
           </RouterLink>
+
+          <button id="show-password-modal-button" class="action-card" @click="showPasswordModal = true">
+            <svg class="action-icon" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="1.5" stroke-linecap="round" stroke-linejoin="round">
+              <rect x="3" y="11" width="18" height="10" rx="2" />
+              <path d="M7 11V7a5 5 0 0 1 10 0v4" />
+            </svg>
+            <span class="action-label">{{t("admin.dashboard.changePassword")}}</span>
```
Simon-VdB: eslint warning

Simon-VdB (changes requested):
Very nicely done.

A few notes:
- I added some suggestions but feel free to argue...
- Currently the only thing needed to reset a password is a valid session cookie. Isn't a proof-of-current-password a bit safer security-wise?
- Also the existing token stays valid after password change. Shouldn't they need to get a new one using the new password?

noahvsb:
> * Currently the only thing needed to reset a password is a valid session cookie. Isn't a proof-of-current-password a bit safer security-wise?
> * Also the existing token stays valid after password change. Shouldn't they need to get a new one using the new password?

Both valid points, for the first one I was too lazy to implement it and the second one went over my head.

noahvsb:
Wait no, nevermind, the session token doesn't include the password and I won't add the password to the token, since that creates a bigger attacking surface. So it's not necessary to change it

But it does include username and super. We don't need the username, but we need the super, so I'll make an issue where you need to check if the token payload corresponds with the database.

noahvsb:
> But it does include username and super. We don't need the username, but we need the super, so I'll make an issue where you need to check if the token payload corresponds with the database.

I think if we need a database fetch anyway, we can drop the super as well.

noahvsb:
Done: #369

noahvsb:
TODO
- [x] add back-end support for old password confirmation
- [x] add front-end support for old password confirmation

noahvsb:
wrong commit message oops, meant to be "update TSDoc"

backend/src/routes/auth/handlers/edit.ts
```diff
@@ -53,4 +59,35 @@ export async function editAdmin(
   );
 
   return parseSchema(server, z.array(AdminSchema), result.rows, ParseContext.Database)[0] ?? null;
+}
+
+/**
+ * Edits the current logged in admin's password and returns an empty 200 OK.
+ *
+ * @param server - The Fastify instance, used for database access and logging.
+ * @param request - The Fastify request, expected to contain a pbody with `password`.
```
Simon-VdB: "pbody" typo?
Also the body now contains oldPassword and newPassword.

```suggestion
 * @param request - The Fastify request, expected to contain a body with `oldPassword` and `newPassword`.
```

backend/src/routes/auth/handlers/edit.ts
```diff
@@ -53,4 +59,35 @@ export async function editAdmin(
   );
 
   return parseSchema(server, z.array(AdminSchema), result.rows, ParseContext.Database)[0] ?? null;
+}
+
+/**
+ * Edits the current logged in admin's password and returns an empty 200 OK.
```
Simon-VdB: 204 No content

```suggestion
 * Edits the current logged in admin's password and returns an empty 204 No Content.
```

frontend/src/services/auth.ts
```diff
@@ -260,6 +260,23 @@ export async function updateAdmin(
...
 
+/**
+ * Update your own password.
+ *
+ * @param password the new password.
+ * @throws {ApiError} 401 — unauthenticated.
+ * @throws {ApiError} 404 — admin not found.
+ *
+ * @example
+ * await updateOwnPassword("hello123");
```
Simon-VdB: this has 2 parameters now...

frontend/src/services/auth.ts
```diff
@@ -260,6 +260,23 @@ export async function updateAdmin(
...
+ */
+export async function updateOwnPassword(
+  oldPassword: string,
+  newPassword: string,
+): Promise<void> {
+  await apiFetch<Admin>(`/auth/me`, { method: "PATCH", body: { oldPassword, newPassword } });
```
Simon-VdB: Maybe this <Admin> is confusing because it returns 204 without body...

backend/src/routes/auth/handlers/edit.ts
```diff
@@ -53,4 +59,35 @@ export async function editAdmin(
...
+  const body = parseSchema(server, EditOwnPasswordSchema, request.body);
+  const { admin } = getMetadata(request);
+
+  // check if old password matches
+  await checkCredentials(server, { id: admin }, body.oldPassword);
+
+  // update with new password
+  const hashedPassword = await hashPassword(body.newPassword);
+
+  // note: I won't update the updated by, since it doesn't change anything to the admin schema or any of the fields visible in the CMS.
```
Simon-VdB, line 85: I didn't think of this in the previous review, but maybe it's necessary for security reasons that an audit-trail is available...

frontend/src/components/admin/ChangePasswordModal.vue
```diff
@@ -0,0 +1,122 @@
...
+const showPassword = ref(false);
+
+function close() {
+  emit("close");
+}
+
+async function submit() {
+  error.value = null;
+
+  if (!oldPassword.value || oldPassword.value.length < 8 || !newPassword.value || newPassword.value.length < 8) {
```
Simon-VdB, line 76: All these scenarios produce the same error message: too short.
The user doesn't know if the error is talking about the new or old password.

Also something subtle to keep in mind. If say an admin was created manually with a password shorter than 8 characters, than said admin would never be able to change his own password because of these checks. I don't know if this is relevant, but I wanted to put it out here so it's in the back of our minds...

backend/src/routes/auth/handlers/login.ts
```diff
@@ -61,4 +45,52 @@ export async function login(
...
+}
+
+type AdminIdentifier =
+  | { username: string }
+  | { id: number };
+
+const fetchAdminCredentials = (server: FastifyInstance) =>
+  buildQuery(
+    server,
+    `SELECT id, password, super FROM admin WHERE username = $1 OR id = $2`,
```
Simon-VdB: 2 query helpers (`fetchByUsername` / `fetchById`) would be more robust.

Simon-VdB (changes requested):
Looks good, 
sorry for the new observations going through the code again.

Something to think about:
Browsers and password managers expect `autocomplete="current-password"` or `autocomplete="new-password"

frontend/src/components/admin/ChangePasswordModal.vue
noahvsb: > All these scenarios produce the same error message: too short.
> The user doesn't know if the error is talking about the new or old password.

I don't think it's relevant, the user can see how long their inputs are. However I think I'll update the error message to explicitly say it should be at least 8 characters.

> Also something subtle to keep in mind. If say an admin was created manually with a password shorter than 8 characters, than said admin would never be able to change his own password because of these checks. I don't know if this is relevant, but I wanted to put it out here so it's in the back of our minds...

It's relevant, because `create-admin.ts` doesn't check the length. I made an issue for this: #373.

backend/src/routes/auth/handlers/edit.ts
noahvsb: > I didn't think of this in the previous review, but maybe it's necessary for security reasons that an audit-trail is available...

I don't think the super admins have any business knowing when an admin changes his/her own password. Also when the super admins see it's been updated recently, they won't think about the password, since you can't see that in the CMS, so they might get worried the admins got access to stuff they should not have access to.

frontend/src/services/auth.ts
noahvsb: I don't see why it's confusing? It returns nothing, because there isn't a body and errors are propagated.

frontend/src/components/admin/ChangePasswordModal.vue
noahvsb: > However I think I'll update the error message to explicitly say it should be at least 8 characters.

I'm stupid, I already did this.

backend/src/routes/auth/handlers/login.ts
noahvsb: I was too fixated on doing it with 1 query, but in the end, it resulted in smelly code :pensive:.

frontend/src/services/auth.ts
Simon-VdB: I meant the `<Admin>`. Could be `<Void>`

Simon-VdB (dismissed):
Thanks for the changes.

Simon-VdB (dismissed):
Looks good, thank you.

noahvsb:
Waiting on #374 to merge this.

Simon-VdB (approved):
Looks good now, let's merge this.

PR #374, opened 2026-04-29
robots.txt on main

noahvsb:
**Issue(s)**

____

**Description**

The robots.txt needs to be on the main branch, so it's hosted on our domain as quickly as possible.

Currently staging doesn't have any functional changes, only dependabot and code refactoring. So besides the robot.txt no functionality will be pushed.
____

**Sources**


Simon-VdB (changes requested):
My thoughts... please argue:

The current setup suggests that robots.txt is not being served correctly on the public domain.

The frontend is served via frontend/nginx.conf with a fallback rule:
`try_files $uri $uri/ /index.html;`
At the same time, the backend runs on a separate container/port, and the API base path (frontend/src/services/api.ts:16) is /api/v1. This implies that the reverse proxy likely only forwards /api/v1/* requests to the backend, while all other requests are handled by the frontend Nginx.
As a result, a request to:
`/robots.txt`
will hit the frontend Nginx, fail to find a matching static file, and fall back to returning index.html (with Content-Type: text/html). This means the actual backend route for robots.txt is never reached.

Search engines will not receive a valid robots.txt, which can negatively affect SEO and crawling behavior.
Suggested fixes:
- Serve robots.txt as a static file directly from the frontend (e.g., place it in the public/static directory), or
- Add an explicit Nginx route to forward /robots.txt to the backend, or
- Ensure the reverse proxy handles /robots.txt separately before the SPA fallback kicks in.

---------------------------------------------------------------

The current robots.txt configuration appears to include:
`Disallow: /`
This directive blocks all web crawlers from accessing any part of the site — including major search engines like Google.
Impact:
The site will not be indexed at all, which is typically undesirable — especially for a newly launched archive site that likely relies on search visibility.

noahvsb:
> The current setup suggests that robots.txt is not being served correctly on the public domain.

Indeed, it doesn't work. I fixed it locally by putting a robots.txt in the public folder, I'll create a PR for this.

> The current robots.txt configuration appears to include: `Disallow: /` This directive blocks all web crawlers from accessing any part of the site — including major search engines like Google. Impact: The site will not be indexed at all, which is typically undesirable — especially for a newly launched archive site that likely relies on search visibility.

That's the whole point of this.

willemDC-Info (approved):
Overall, a great improvement to the CI/CD setup—deployments are now much more consistent and easier to manage.

One small remark (non-blocking):
*Node 24 is quite new and could potentially cause dependency issues. If that happens, we could consider switching to a stable LTS version (e.g. 20).*

Overall, these changes make the pipeline more reliable and maintainable.

**Approved 👍**

Simon-VdB (approved):
Looks good, requested changes have been done or mitigated. I approve.

Acher0s (approved):

PR #383, opened 2026-05-01
serve static robots.txt file in frontend/public/ directory

noahvsb:
**Issue(s)**

____

**Description**

Our current robots.txt setup doesn't work, I just went for the simplest solution and added a static robots.txt file to the public directory.
____

**Sources**


Simon-VdB (approved):
Thank you for the changes, looks good.

PR #384, opened 2026-05-02
add password validation to hashPassword

### LINKED ISSUE #373: [DEV] check password length in `hashPassword` function
**Location**

`/backend/src/routes/auth/handlers/hash.ts`

____

**Description**

When creating or editing an admin, zod checks if the password in the body is greater than 8 characters. But when inserting the first admin with the `create-admin.ts` script, there's no zod checking. However instead of checkking the length in that script, I would just check the length in the `hashPassword` function, since that one is used on all passwords, for extra safety. We can't really enforce password length in the database since that only includes hashed passwords with fixed length.

____

**Sources**

noahvsb:
**Issue(s)**

#373 
____

**Description**

Title speaks for itself and the reason of the validation is inside the [issue description](https://github.com/SELab-2/viernulvier-1/issues/373#issue-4348896617).

You can test it manually by using the `create-admin.sh` script. That script previously didn't have any password validation.
____

**Sources**


backend/src/routes/auth/handlers/hash.ts
```diff
@@ -1,17 +1,30 @@
+import { ValidationError } from "@/routes/helpers.js";
 import bcrypt from "bcrypt";
+import z from "zod";
 
 const SALT_ROUNDS = 12;
 
+const PasswordValidationSchema = z.object({
+  password: z.string().min(8).max(72),
```
Simon-VdB: this rule now lives in 4 places create.ts, edit.ts (where it's already extracted as zPassword), replace.ts and now hash.ts. 
Since this PR is explicitly about centralizing the rule, it would be a good moment to export a single shared zPassword (or PasswordSchema) and have all four call sites import it.

Simon-VdB (changes requested):
Sorry for the requested change, it's just time to refactor a 4 times duplicate code fragment.

backend/src/routes/auth/handlers/hash.ts
noahvsb: The `zPassword` extraction is on #385. It will create merge conflicts, I'll do it after merging that PR.

noahvsb: No wait it's not, I can't find the `zPassword`, I probably need to merge staging.

Simon-VdB (approved):
Thank you. Nicely done.

Issue #1, opened 2026-02-16
[DESIGN] EER-model of database


noahvsb:
**File**

____

**Description**

Create an EER-model for the database.

PR #385, opened 2026-05-02
fix consistency between session token payload and database

### LINKED ISSUE #369: [BUG] fix consistency between session token payload and database.
**Description**

The session token encodes a bit of information like id, username, super. But when an admin is changed and the session token remains the same, there's wrong information in there. However you can't do an API call, because you need a session token to do this. So an SQL query is probably needed, but when you're fetching the data anyway, you don't need to encode it in the session token. The person implementing this has some freedom because I don't know what the best approach is currently.
____

**Environment**

all

**Sources**

`/backend/src/plugin/authorize.ts`, `/backend/src/plugin/jwt.ts`, `/backend/src/routes/auth/handlers/login.ts`

noahvsb:
**Issue(s)**

#369 
____

**Description**

Decided to go with an SQL query to fetch the data from the SQL.

Big problem with the tests:
Since the tests all use a mocked instance of `server.pg.query`, the tests need to also mock a select query on the admin. There are 56 failing test files :saluting_face:

- [x] fix tests

____

**Sources**


noahvsb:
For a change with 17 additions and 8 deletions, I needed 327 additions and 5 deletion to the test files :saluting_face:

noahvsb:
Changed 4 source files and 62 test files :sob:

backend/src/plugins/authorize.ts
Simon-VdB, line 1: 2) re-implements the real plugin (jwtVerify, denylist, super check). Now there are two implementations that must be kept in sync — if someone adds rate-limiting, scope checks, audit logging, etc. to the real plugin, the mock will silently diverge and tests will pass while production differs. Cleaner approach would be to inject the admin lookup as a dependency (so tests can stub just that one thing) rather than mocking the whole plugin.

backend/src/plugins/authorize.ts
```diff
@@ -19,13 +20,21 @@ export default fp(function authorizePlugin(server: FastifyInstance) {
...
 
-        const payload = request.user as { jti?: string; super?: boolean };
+        const payload = request.user as { id?: number, jti?: string };
 
         if (payload.jti && server.tokenDenylist.has(payload.jti)) {
           return await reply.status(401).send({ error: "Token has been revoked" });
         }
 
-        if (options.super && !payload.super) {
+        if (!payload.id) throw Error();
```
Simon-VdB: 4) Bare errors with no message. The catch swallows it without logging anything. At minimum: distinct messages and a server.log.warn(err) in the catch.

Also: standard JS gotcha: !0 is true. Postgres SERIAL starts at 1 so it works today, but this is a fragile pattern. Prefer payload.id === undefined or typeof payload.id !== "number".

```suggestion
        if (payload.id === undefined) throw Error();
```

backend/src/plugins/authorize.ts
Simon-VdB: stale doc comment
The doc should describe the new behavior (live DB lookup of the admin's super flag).

Simon-VdB (changes requested):
Overall: the security/correctness intent is right, but point [#1](https://github.com/SELab-2/viernulvier-1/issues/1) deserves a real conversation before merge — the PR introduces a per-request DB query that the rest of the system isn't designed to absorb. Points [#2](mock divergence), [#3](missing regression test), and [#4] (silent errors) are also worth addressing.

1) This is the most important point. Previously authorize only verified the JWT signature (a CPU-only operation). After this PR, every authorized request fires an extra SELECT … FROM admin WHERE id = $1 — not just super-only routes. The result is then thrown away (the admin is fetched again later by handlers like fetchCurrentlyLoggedInAdmin).

The intent is correct (so demoting/deleting an admin takes effect before the 24h JWT expires), but the cost is real:

- Adds DB latency to every API call.
- Doubles or triples DB query load per request.
- Becomes an amplification vector under high traffic.
- Makes auth a hard dependency on DB health (transient DB blips → users get logged out).
Worth discussing alternatives before merging:

- Short-lived access tokens + refresh tokens (e.g., 5–15 min access, refresh re-checks DB).
- Invalidate tokens by jti on permission change — already have tokenDenylist, just add the user's jti to it when their super flag flips.
- Attach the fetched admin to request so handlers can reuse it (avoids the double query).
- Short-TTL in-memory cache (e.g., 5–10 s) keyed by admin id.
At minimum the fetched admin should be reused by downstream handlers.

---------------------------------------------------------------------------------------

3) The PR's whole purpose is "make permission checks reflect current DB state, not the stale JWT claim". The test suite covers "admin not found → 401" and the existing super/non-super cases, but not the core scenario: admin still exists, super flag was flipped after the token was issued, the next request should reflect the new value. Adding that test would protect the fix from being regressed by point 1's eventual cache layer.

noahvsb:
> 1. This is the most important point. Previously authorize only verified the JWT signature (a CPU-only operation). After this PR, every authorized request fires an extra SELECT … FROM admin WHERE id = $1 — not just super-only routes. The result is then thrown away (the admin is fetched again later by handlers like fetchCurrentlyLoggedInAdmin).
> 
> The intent is correct (so demoting/deleting an admin takes effect before the 24h JWT expires), but the cost is real:
> 
> * Adds DB latency to every API call.
> * Doubles or triples DB query load per request.
> * Becomes an amplification vector under high traffic.
> * Makes auth a hard dependency on DB health (transient DB blips → users get logged out).
>   Worth discussing alternatives before merging:
> * Short-lived access tokens + refresh tokens (e.g., 5–15 min access, refresh re-checks DB).
> * Invalidate tokens by jti on permission change — already have tokenDenylist, just add the user's jti to it when their super flag flips.
> * Attach the fetched admin to request so handlers can reuse it (avoids the double query).
> * Short-TTL in-memory cache (e.g., 5–10 s) keyed by admin id.
>   At minimum the fetched admin should be reused by downstream handlers.

This really isn't that much of a problem, if we're lucky we'll get 3 admins working on it at the same time. I think it's worth it for the extra security. There are other solutions, but I don't think they're worth it to implement.

backend/src/plugins/authorize.ts
noahvsb: I think this comment is on the wrong file, I think it should have been on `backend/test/__mocks__/plugins/authorize.ts`.

Yes they should be kept in sync, but I don't really see the problem with it. The tests might pass, but the coverage threshold won't, so you won't accidentally push untested code. However I'll add a comment to the plugin that the mock should be kept in sync to make it easier for future devs to work on this.

Your cleaner approach was one I tried before and it might be a bit cleaner for tests in the long run, but it was a pain in the ass to implement, I had to change way more to the tests and it would've probably taken me an extra day of work. So I went with the easier option to implement, which is good enough in my opinion.

noahvsb:
> 3. The PR's whole purpose is "make permission checks reflect current DB state, not the stale JWT claim". The test suite covers "admin not found → 401" and the existing super/non-super cases, but not the core scenario: admin still exists, super flag was flipped after the token was issued, the next request should reflect the new value. Adding that test would protect the fix from being regressed by point 1's eventual cache layer.

I added another test to `backend/test/routes/auth/integration.test.ts`.

backend/src/plugins/authorize.ts
noahvsb:
```sh
viernulvier-backend      | {"level":30,"time":1778323115691,"pid":318,"hostname":"317a4810ae07","reqId":"req-3","req":{"method":"GET","url":"/api/v1/auth","host":"backend:3000","remoteAddress":"172.18.0.5","remotePort":53744},"msg":"incoming request"}
viernulvier-backend      | {"level":30,"time":1778323115691,"pid":318,"hostname":"317a4810ae07","reqId":"req-3","res":{"statusCode":401},"responseTime":0.3239420000463724,"msg":"request completed"}
```
The server already logs all requests and reponses, I don't think an extra log is necessary.

Simon-VdB (approved):
Nicely done, thank you for the requested changes. See, I manipulated you into changing one more file so we would have 67 files changed...
I approve this very much sir. Let's merge

PR #391, opened 2026-05-03
Productions page loading placeholders

### LINKED ISSUE #299: [DEV] productions page loading animation
**Location**

`frontend/src/views/ProductionsView.vue`

____

**Description**

Create a loading animation for the productions page.

____

**Sources**

https://jamesinkala.com/blog/make-animated-content-placeholders-with-html-and-css/

trherber:
**Issue(s)**

Closes #299 

____

**Description**

I created pulsing placeholders (skeleton states) for the production cards that are shown while the productions are loading. There is still a layout shift caused by the filters being fetched. A skeleton state could also be implemented for this but I haven’t done that yet since we are still planning to adjust the filter layout anyway.
I’ll create an issue for the filter layout update and the corresponding skeleton state.

The skeleton state is only used on the initial page load. For filter changes skeletons caused too much flicker so I switched to a lighter loading state where existing productions are slightly faded to indicate new data is loading without a disruptive visual change.
____

**Sources**

### DIFF
--- a/frontend/src/components/productions/ProductionListCard.vue
+++ b/frontend/src/components/productions/ProductionListCard.vue
@@ -5,7 +5,8 @@
       params: { lang: locale, id: production.id },
     }"
     class="production-list-card group -mx-3 flex items-stretch gap-4 border-b border-surface-3 px-3 py-8 transition-colors last:border-b-0 hover:bg-surface-1/60 sm:-mx-4 sm:gap-6 sm:px-4 md:-mx-5 md:gap-8 md:px-5"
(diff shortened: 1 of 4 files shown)


lucasodb:
If we're doing skeletons, then we should do it for the whole page I think, so also for the filters. This would kind of have to be hardcoded though, and the layout of the filters section is not final yet, so I would keep this pr open for a while and implement full skeletons for this page when we're sure of the layout (or maybe decide to leave this out altogether).

lucasodb:
Another pr was made for this, so this one can be closed.

PR #400, opened 2026-05-04
Remodel filter layout

lucasodb:
**Issue(s)**

____

**Description**

General visual and user experience improvements to the filter system on the productions page. Previously this was way too cluttered and noisy, there is now a collapsable filter panel instead for the tags and genres, and by default only a single genre row is shown. The date filter was moved up, next to the search bar. I plan to also add a sort button (date/name) to the right of the search bar, in place of the current search button, but since this requires some backend work as well I'll probably do this in a seperate pr.

If you review this please wait with merging, I want to merge #407 first so I can properly intergrate this into this pr.

____

**Sources**


noahvsb (changes requested):
I haven't reviewed the code yet, I'll do that after #407 has been merged, because I expect some changes.

The date remodel looks good.

For the search bar, I do expect the email-like remodel as we discussed in the meeting, which isn't yet implemented.

For the tags I had 2 issues:
- The button to expand the tags component is on the left, but the button to collapse is on the right. I think it's best to have both be on the same side.
- The "+" button to expand the the tags of a specific tag type can be confused with a button to create a new custom tag. I think the previous version with "Show more" and "Show less" is better. You could also do "X more..." to indicate how many more tags.

lucasodb:
> I haven't reviewed the code yet, I'll do that after #407 has been merged, because I expect some changes.
> 
> The date remodel looks good.
> 
> For the search bar, I do expect the email-like remodel as we discussed in the meeting, which isn't yet implemented.
> 
> For the tags I had 2 issues:
> 
>     * The button to expand the tags component is on the left, but the button to collapse is on the right. I think it's best to have both be on the same side.
> 
>     * The "+" button to expand the the tags of a specific tag type can be confused with a button to create a new custom tag. I think the previous version with "Show more" and "Show less" is better. You could also do "X more..." to indicate how many more tags.

I won't change the search bar here too much, because that's not what this pr is for, so I'll do this in a seperate pr. Also, the expand button is not mean to fall of the row but always be anchored to the right, I thought that was the case but seems like it's not always so. Could you maybe send a screenshot of what this looks like? I supposed 1 more tag pill is displayed on the row than actually fits, which means `useFittingPills` is doing something wrong. This is what it looks like for me:
image
Please try to set `trailingControlGapPx` to 8 instead of 4 in this section:
```js
const {
  setRowRef: setCompactGenreRowRef,
  setTrailingControlRef: setCompactGenreExpandButtonRef,
  setPillRef: setCompactGenrePillRef,
  visibleItems: compactGenreTagsForRow,
} = useFittingPills(compactGenreTagCandidates, {
  gapPx: 8,
  trailingControlGapPx: 4,
  fallbackVisibleCount: COMPACT_GENRE_FALLBACK_MAX,
});
```
If that doesn't fix the problem, it might be that the spacing is not consistent across browsers or something, I'll have to look into it.

I agree that the `+` might be confusing.
I changed this from text, because with text, 'Minder tonen' for example is longer than 'Meer tonen', which causes a layout shift in the tag pills, which doesn't look good. An icon of a fixed size solves that issue, but I agree that this isn't clear enough, so I'll try to think of something else.

noahvsb:
image

The `gapPx` was already 8 when I tested this.

Here's the result after expanding:

image

Note: I hadn't pulled the last 2 commits yet, now I have and the issue stays

lucasodb:
> image
> 
> The `gapPx` was already 8 when I tested this.
> 
> Here's the result after expanding:
> image
> 
> Note: I hadn't pulled the last 2 commits yet, now I have and the issue stays

I asked to change `trailingControlGapPx`, not `gapPx`. I hadn't attempted to fix this problem yet in these two commits, I first want to find out what causes it. I assume this change won't fix it anyways.

lucasodb:
I didn't intend for this to be a general fixes pr, but since I noticed several issues after the merge, I will just make changes that aren't fully relevant to this pr in order to not push broken code. This includes a fix to the sorting, fixing the visuals of the date slider on chrome, and the displaying of active search queries.

frontend/src/views/ProductionsView.vue
noahvsb, line 1: This file is getting way too big, can you make an issue for refactoring this file?

frontend/src/views/ProductionsView.vue
```diff
@@ -1145,30 +1210,76 @@ const nonGenreTagsForFilter = computed(() => {
...
+);
+
+const {
+  setRowRef: setCompactGenreRowRef,
+  setTrailingControlRef: setCompactGenreExpandButtonRef,
+  setPillRef: setCompactGenrePillRef,
+  visibleItems: compactGenreTagsForRow,
+} = useFittingPills(compactGenreTagCandidates, {
+  gapPx: 8,
+  trailingControlGapPx: 4,
```
noahvsb:
```suggestion
  trailingControlGapPx: 8,
```

This did fix the expand button. It's now on the right as expected.

noahvsb (changes requested):
Visually it looks good now.

frontend/src/views/ProductionsView.vue
lucasodb: #424

noahvsb (approved):

Issue #401, opened 2026-05-05
[DEV] UI/UX enhancement in production detail page


Simon-VdB:
**Location**

`/frontend/src/views/ProductionDetailView.vue`
____

**Description**

Make the UI more archive like.

____

**Sources**

N/A

PR #402, opened 2026-05-05
Feat/archive feel production detail

Simon-VdB:
**Issue(s)**

https://github.com/SELab-2/viernulvier-1/issues/401

____

**Description**

Enhanced the production detail page UI to look more archive like.

____

**Sources**

N/A


lucasodb:
Some points:
- I like the font change, looks way better. We should consider consistency if we use this, and setup rules to use this everywhere, so our fonts aren't all over the place .
- I think the run period and running time should definitely be displayed as well. You kept only the year.
- I think all font sized in the card should be a bit larger, but especially artist is super small and all the way at the bottom, this is an important field and should better be displayed above the title.
- 'Image from the archive of De Vooruit' shouldn't be displayed below the card, there isn't really any use for this.
- Very minor, but the 'tags' title for the tags panel was bold previously, I think it should be here too, and maybe the tags themselves should stand out a little bit more as well, because they don't really right now.
- This is shown below descriptions for some reason, not sure what this is?
image

- The quotes on a quote are only added on one side
image

---
I really like the font and new layout of the description text, I think the card has potiential too, keep trying at it.

Simon-VdB:
> * I like the font change, looks way better. We should consider consistency if we use this, and setup rules to use this everywhere, so our fonts aren't all over the place .

That's why I created PR https://github.com/SELab-2/viernulvier-1/pull/404 : it changes DESIGN.md in a way that is not ambiguous and deterministic so everyone that uses it, has the same look and finish.

Simon-VdB:
Thank you for the review, I've made some changes that you suggested. Let me know what you think.

lucasodb:
Looks better.

I'm not really a fan of this bar below the banner image. Because we show images across the whole screen of resolution 1920x600, this takes up a very large part of the screen already. Having this bar in addition below the image, makes it so you can most of the time basically see nothing of the other info for the production on the first screen. We don't have this problem currently because all the header text is displayed on top of the image itself. 

I think the info card should be displayed a bit higher, so this empty bar below the header is much smaller. If it's too high it will of course sometimes block an important part of the image, like a face, so that would not be ideal. You could also make the card expand a bit more horizontally instead of only vertically, so the height of the bar doesn't get worse the larger the title etc. is

Of course, making the image smaller is also a solution, but I think we've committed to the full screen image esthetic so maybe we should stick to it. We could however just chop of something like 10-20% of the top of the image while keeping it at the same scale, so only cropping the image without making it zoom in or stretch more, I would definitely consider that.

When you apply any of these changes, please do it in seperate commits instead of a single one, that is a bit easier for me to go over all the changes you made one by one.

image
image

frontend/src/components/production/HeroSection.vue
```diff
@@ -1,9 +1,28 @@
...
+      kicker  →  artist  →  title  →  deck  →  run / duration strip
+
+    The artist is the primary identifier in a concert / theatre
+    archive (you remember "I saw The Cure", not "I saw the Faith
+    Tour") so it sits above the title in italic serif. The strip
+    underneath surfaces the run period and running time at a glance —
+    the same facts also live in EventsSection, but readers expect to
+    see them at the top.
+  -->
+  <article class="relative bg-surface-0">
```
lucasodb:
```suggestion
  <article class="relative bg-surface-1">
```

frontend/src/components/production/HeroSection.vue
```diff
@@ -13,76 +32,68 @@
...
+        >
+          {{ dateMetaLine }}
+        </p>
       </div>
-    </div>
-  </section>
+    </header>
+
+    <!-- Breathing room before the next section -->
+    <div class="h-12 md:h-16" aria-hidden="true" />
```
lucasodb:
```suggestion
```

lucasodb (changes requested):
These changes make it so there isn't a section visibly dividing the banner and the detail section. I also noticed that the font is applied almost everywhere, but the event's halls and their address still use the previous font, maybe we should also change this to another font, you can disagree with me on this. Lastly, I would also remove the comments in the templates on the DetailSection and Herosection for consitency, and because they are not really needed. Or, keep it just to "Editorial header for a production page." and "Sidebar — marginalia first, then a tag drawer below." without the rest of the text.

lucasodb (approved):

PR #404, opened 2026-05-05
Rewrite DESIGN.md as a binding system for both site registers

Simon-VdB:
**Issue(s)**

https://github.com/SELab-2/viernulvier-1/issues/401

____

**Description**

The old DESIGN.md was largely a colour palette plus a few brief notes on buttons and typography, written when the landing page was still the only reference. Most of the choices we made while polishing the production detail page (kadertje, drop cap, pull quote, marginalia, museum-label headings, credits footer, end-mark, the serif/sans pairing, the photographic dual register) lived nowhere in the docs and would have to be re-derived for every new page.

This commit replaces the file with a full system spec, organised so a developer can pick a new page, decide which register it lives in, and build it without inventing patterns.

What's new:

- §0 names the two registers explicitly (editorial article vs catalogue / listing) and ties every later pattern to one of them.
- §4 codifies the Source Serif 4 + Inter pairing, the type scale, and a forbidden list (`font-black`, `tracking-tighter`, italic on UI, uppercase on body, font-mono on prices).
- §5 splits photography into editorial (colour, no filter) and archival (grayscale + contrast-125), with concrete classes and forbidden cases.
- §6 introduces the editorial article patterns: kicker, kadertje, drop cap (with `initial-letter` + float fallback), pull quote, asterism break, photo dateline, marginalia, museum-label heading, credits footer, end-mark, body conventions.
- §7 introduces the catalogue patterns: section heading shared with editorial h2, list rows, tag chips (two variants only — solid genre
  + outline generic, both neutral), tag drawer, card.
- §8 nails buttons to three variants — primary, ghost, hairline-link inline — and forbids coloured fills, rounded-full, trailing arrows on non-paginated CTAs.
- §10 documents page-level rhythm separately for article and catalogue pages, and lists the patterns forbidden on every page (bento grid, marketing CTAs, italic/uppercase/font-black titles, concert-poster heros).
- §14 collapses everything into a thirteen-rule quick reference for PR review.
- §15 points reviewers at the canonical implementations (HeroSection / DetailsSection / ProductionListCard).

The file is aspirational where the codebase has not caught up yet: the production detail PR (feat/archive-feel-production-detail) implements §6 in full, but the home and productions list pages still use the old shouty patterns. They should be brought in line as follow-up work.

No code changes; documentation only..

____

**Sources**

https://github.com/SELab-2/viernulvier-1/pull/402


lucasodb (approved):
This can be helpful as a guideline, and for refactoring pages to match the current detail page esthetic, but should probably not be seen as a binding guideline. Most of the items here are very use case specific and don't really work in other context per se, but that's fine. It says for example that tag chips should never be fully round, and that might look good for the ones on the detail page, but I don't think the tags pills on the productions page should look like that per se. Buttons is another area where you shouldn't try to be super strict, because it's a very versatile component, and is used in different ways on the productions page. But for setting standards for fonts and color schemes etc., this is useful.

PR #407, opened 2026-05-07
Add sort functionality

lucasodb:
**Issue(s)**
____

**Description**

Added option to sort results on name and date. The queries for this will also be added to the URL. I added an index for efficient lookup.

____

**Sources**


frontend/src/components/productions/ProductionsSortControl.vue
noahvsb, line 1: I think it will be cleaner to put a lot of the css in `<style>` instead of inline.

frontend/src/views/ProductionsView.vue
noahvsb, line 1: More of the styling inside the `<style>` instead of inline css would be nicer here as well.

noahvsb (changes requested):
Looks good and it's very fast for me (I do have to note that I only have the legacy productions).

Just 2 small things regarding cleaner code:

noahvsb (approved):
Well done, much cleaner now 👍

PR #414, opened 2026-05-08
fix: docker build

noahvsb:
**Issue(s)**

____

**Description**

The docker build in the staging branch was failing. I fixed it by pinning the `pnpm` version in the `Dockerfile`s and `adding --ignore-scripts` to the frontend `Dockerfile`.

I tested it locally by running these 2 commands:

```sh
# backend
docker buildx build --file ./backend/Dockerfile --platform linux/amd64 .

# frontend
docker buildx build --file ./frontend/Dockerfile --platform linux/amd64 .
```
____

**Sources**

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -20,7 +20,7 @@ COPY --from=deps /app /app
 
 COPY . .
 
-RUN pnpm install --ignore-scripts
+RUN pnpm install
 RUN pnpm --filter frontend build
 
 
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -1,6 +1,6 @@
 FROM --platform=$BUILDPLATFORM node:24 AS base
 
-RUN npm install -g pnpm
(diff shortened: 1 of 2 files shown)


frontend/Dockerfile
```diff
@@ -20,7 +20,7 @@ COPY --from=deps /app /app
 
 COPY . .
 
-RUN pnpm install
+RUN pnpm install --ignore-scripts
```
Acher0s: this could have impact on OS specific dependency install scripts. is it truly necessary? it could become a headache later if we end up adding more dependencies, or maybe it won't work in our production environment (i'd have to test it).

Acher0s (changes requested):

frontend/Dockerfile
noahvsb: Just tested and it's not necessary. I tried a few things and then stripped down a lot after getting it working, but thought it was necessary.

Acher0s (approved):

PR #416, opened 2026-05-09
fix dev docker build

noahvsb:
**Issue(s)**

____

**Description**

Same issue and same fix as the non-dev dockers that I fixed yesterday: #414. I missed these yesterday because I thought they worked (I never tested).

____

**Sources**


Simon-VdB (approved):
Yes, very handy and critical issue fix. I approve

PR #421, opened 2026-05-11
Updated blogpost types to languagemap

trherber:
**Issue(s)**

____

**Description**

I changed the types of title content, name and description in blogpost and blog to languagemap. I’m making this small PR now so anyone else planning to work with blogposts can use the same types.

____

**Sources**


noahvsb (approved):
Code looks good, migrations work, tests succeed.

Issue #425, opened 2026-05-11
[BUG] fix <details> drop-cap too many times


Simon-VdB:
**Description**

The dropcap should only happen on the first description.

____

**Environment**

N/A

**Sources**

<img width="2590" height="1366" alt="Image" src="https://github.com/user-attachments/assets/a383b21f-df16-4768-b432-bb725d27953d" />

PR #426, opened 2026-05-11
fix(production-detail): scope drop-cap to direct first child

Simon-VdB:
**Issue(s)**

https://github.com/SELab-2/viernulvier-1/issues/425

____

**Description**

The `.article-lead :first-child::first-letter` descendant selector matched any element that happened to be a first child inside the lead container, including an `<a>` that is the only child of a following `<p>`. This caused a second drop cap to appear (e.g. the "K" of "Klik hier..."). Use a direct-child combinator so only the lead's first direct child is targeted.

____

**Sources**

N/A

### DIFF
--- a/frontend/src/components/production/DetailsSection.vue
+++ b/frontend/src/components/production/DetailsSection.vue
@@ -243,16 +243,19 @@ const mainContentClass = computed(() => {
 }
 
 /*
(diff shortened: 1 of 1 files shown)


lucasodb (approved):
Good fix.

PR #429, opened 2026-05-11
Batch operation for fetching productions media

### LINKED ISSUE #390: [DEV] Batch operation for fetching production images
**Location**

Media endpoint, `ProductionsView.vue`

____

**Description**

Use a batch operation to fetch media for the productions instead of parallel requests as is done now in `loadThumbnailsForProductionIds` of `ProductionsView.vue`.

____

**Sources**

lucasodb:
**Issue(s)**

Closes #390 

____

**Description**

Implemented a batch operation for fetching productions media on the productions page, instead of making individual parallel requests, as suggested by @Simon-VdB in #365 

The scroll to top animation now feels a bit buggy and inconsistent on firefox I've noticed, on chrome it feels great, so this might need to be done differently but there already exists and issue for this.

____

**Sources**

### DIFF
--- a/frontend/src/services/media.ts
+++ b/frontend/src/services/media.ts
@@ -9,6 +9,12 @@ import { apiFetch, ApiError } from "./api";
 
 export type ImageWithCrops = Image & { crops: Crop[] };
 
+export type ProductionImagesBatchResponse = {
+  byProductionId: Record<string, ImageWithCrops[]>;
+};
+
+const MAX_IDS_PER_PRODUCTION_IMAGE_BATCH = 50;
+
 /**
  * All images (with nested crops) for a production.
  *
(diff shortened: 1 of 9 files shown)


frontend/src/views/ProductionsView.vue
```diff
@@ -972,7 +973,7 @@ async function fetchProductionsPageData(page0: number) {
   syncFilterBannerFromApplied();
   eventsByProduction.value = eventsMap;
   thumbnailUrlByProductionId.value = new Map();
-  void loadThumbnailsForProductionIds(ids);
+  await loadThumbnailsForProductionIds(ids);
```
Simon-VdB: fetchProductionsPageData now only resolves after all thumbnails have been loaded. This means the loading spinner on the productions list remains active until the batch fetch (including all crop joins) has completed, whereas previously the list became visible immediately and thumbnails loaded asynchronously afterward.

The thumbnailLoadGeneration counter exists specifically to cancel stale fire-and-forget loads — it is now less relevant. Whether this is the desired UX is a product decision, but it is a deliberate behavioral change that may make the page feel slower on slow connections.

frontend/src/services/media.ts
```diff
@@ -54,3 +60,66 @@ export async function getImagesForProductionOrEmpty(
     handleProductionImagesFetchError(productionId, err),
   );
 }
+
+function handleProductionImagesBatchFetchError(
+  productionIds: readonly number[],
+  err: unknown,
+): void {
+  if (err instanceof ApiError && err.status === 404) {
+    return;
```
Simon-VdB: completely silent.
The batch endpoint /production/images never returns a 404 — it always returns 200 with an empty map. A 404 here means the route does not exist (e.g. due to a deployment mismatch or incorrect prefix). Silently swallowing that hides real configuration issues.

At minimum, a console.warn would be appropriate here, just like for the other status codes.

Simon-VdB (changes requested):
Nicely done. Thank you for the implementation, it looks solid, here are some things to discuss.

mediaRoutes was moved before productionRoutes. The likely reason was to avoid conflicts between /production/images and /production/:productionId/image, but those routes have different URL depths (4 vs. 5 segments) and therefore never conflict in Fastify’s radix-tree router.
The order does not matter for correctness. A short comment explaining the ordering would help future developers — otherwise someone may incorrectly conclude that the order is semantically important.

lucasodb:
> Nicely done. Thank you for the implementation, it looks solid, here are some things to discuss.
> 
> mediaRoutes was moved before productionRoutes. The likely reason was to avoid conflicts between /production/images and /production/:productionId/image, but those routes have different URL depths (4 vs. 5 segments) and therefore never conflict in Fastify’s radix-tree router. The order does not matter for correctness. A short comment explaining the ordering would help future developers — otherwise someone may incorrectly conclude that the order is semantically important.

The order does actually matter. I changed it because it caused problems. I'll add a comment explaining why.

Simon-VdB (approved):
This is clear now, thank you.

PR #431, opened 2026-05-12
Fix/integration tag production

### LINKED ISSUE #389: [BUG] Production endpoint doesn't update tags
**Description**

There is currently no endpoint on production to change the tags assigned to a production.
On top of that, the tags should be created at the same time as a production is created.

____

**Environment**

Production endpoint routes

msth5:
**Issue(s)**

closes #389 

____

**Description**

Integrated tag assignment to production endpoints and vice versa.
Left the old endpoint in place as they're still in use.


Acher0s (approved):
didn't review the tests but assume theyre fine

PR #438, opened 2026-05-13
Display productions as 3-column grid instead of list

### LINKED ISSUE #437: [DEV] Display productions as 3-wide grid instead of list
**Location**

`frontend/src/views/ProductionsView.vue`, `frontend/src/components/productions/`

____

**Description**

On the productions page, replace the long vertical list of productions with a 3-column grid of cards. All the information currently displayed on a production card (title, artist, dates, halls, tag chips, thumbnail) should remain, but the layout can be reorganised to fit the new card shape.

Decided in meeting on 2026-05-13.

____

**Sources**

Meeting notes 2026-05-13

Simon-VdB:
## Summary
- Replace the productions-page list with a responsive 3-wide grid (1/2/3 columns at sm/md/lg); all info from the former list card (title, artist, dates, halls, tag chips, thumbnail) is kept but reorganised vertically for the card shape.
- Widen the cards section to `max-w-7xl` with tighter horizontal padding so the grid fills the page; bump page size from 20 to 21 for a clean 7 × 3 grid.
- Rename `ProductionListCard` → `ProductionGridCard` (file + test), update `ProductionsView.test.ts` limit/offset assertions.

Closes #437

## Test plan
- [x] `npx vitest run` (frontend) — all 1084 tests pass locally
- [x] Visit `/nl/productions` with real data: cards display 3 across on desktop, 2 on tablet, 1 on mobile
- [x] Pagination shows 21 per page; page 2 fetches `offset=21`
- [x] Title, artist, date summary (incl. "& n more"), halls and tag chips all render on each card
- [x] Empty-state / loading-state / error-state still display correctly
- [x] Dark mode card still looks right


frontend/src/views/ProductionsView.vue
```diff
@@ -539,7 +543,7 @@ import {
   tagMapById,
 } from "@/utils/productionsOverview";
 
-const PAGE_SIZE = 20;
+const PAGE_SIZE = 21;
```
noahvsb:
```suggestion
const PAGE_SIZE = 30;
```

Use 30 instead of 21 so it's a round number.

Before:

Image

After:

Image

You can't deny that it looks way cleaner.

On my current desktop it's still instant loading, so performance shouldn't be an issue.

noahvsb (changes requested):

noahvsb (dismissed):
It's good for me now, but I see lucas has self-requested a review, so I won't merge yet.

lucasodb:
Hmm, this is not really what I had in mind. The grid is not meant to replace the list view, but to be a second layout option, so the user can choose themselves and switch between them. We could put a small list/grid layout switch button to the right of the sort component. I was planning to not show all the info on the cards that is shown usually on them, like leaving out the halls and only showing a few tags, because otherwise they can get way too tall, and since all the cards in a row get the same height, this can look pretty bad. I think the thumbnails are also too square, they can just keep the same or similar shape as on the normal list cards, I think that would look better, and you would really be able to see a ton of productions at once, which is the whole goal of the grid view. Sorry if this is a bit harsh, but that's why I suggested doing this myself, because I've had this idea for a few weeks now and I've had some time to consider these things. Before you make any changes though, it's probably best to wait before the productions page refactor has been merged, so you can pull staging and start from there. If you prefer I could also take this over and start over, since this is not a very big change, it's up to you.
image

lucasodb:
Please merge staging before you continue.

frontend/src/views/ProductionsView.vue
```diff
@@ -37,7 +37,14 @@
         </div>
       </section>
 
-      <section class="mx-auto max-w-[82rem] px-6 pb-20 pt-10 lg:px-10">
+      <section
```
lucasodb: We want the width of the page to always remain the same, so undo this change you did and go back to ` <section class="mx-auto max-w-[82rem] px-6 pb-20 pt-10 lg:px-10">`

frontend/src/views/ProductionsView.vue
```diff
@@ -558,11 +588,33 @@ import {
...
 } from "@/utils/productionsOverview";
 
-const PAGE_SIZE = 20;
+const PAGE_SIZE = 30;
 
 /** Same cap as the list API, extra terms are ignored client-side. */
 const MAX_SEARCH_TERMS = 20;
 
+/** localStorage key persisting the user's preferred layout across visits. */
+const LAYOUT_STORAGE_KEY = "productionsLayout";
```
lucasodb: I really don't think we should save the preferred layout, that's just not necessary and can only cause problems, so I would remove this logicl.

frontend/src/components/productions/ProductionGridCard.vue
```diff
@@ -0,0 +1,180 @@
+<template>
```
lucasodb, line 1: You have a totally different style now for the grid cards then for the list cards? I asked you to pull from staging, the whole point was of course to use the new styling that was chosen for the list cards, but now all the old styling is still used here. You should be able to reuse most things on the list card here.

lucasodb (changes requested):
This is still not great. Keep the same page width and layout for both views. You can just keep what it was. Also use the new refactored styling we did for the productions page for the grid cards. Make the thumbnails the same shape as they are on the list cards as well, you can make them larger but make the shape be the same.

lucasodb (changes requested):
I have a few more remarks. I think it would look better if there was less variation in the card sizes. To achieve this, we could we can make it so the the title and the artist are always only 1 line long, and we just show an ellipsis, instead of showing the full title, when it's longer then the one line. We can also only show as many tag pills as fit in 1 row, instead of a hardcoded amount. Then you can also leave out the "moreGridTags" part. For the collapsed genre row I used `useFittingPills`, maybe you can reuse that logic, I'm not sure if that would translate 100% though. I would also like to see the images be a bit bigger, so less padding around the thumbnail and the entire card in general, just use thinner edges, I think that would look better. I also think the highlight that shows which layout is selected can be a bit more subtle, using the inverse color really makes it stand out more than anything else on the page, which is not really the goal. I also think 30 productions on a page is a bit too much, maybe change it to 18.

lucasodb:
One more thing, why did you have to redefine the two placeholder images and the map pin icon, where these not already present on staging?

Simon-VdB:
Those were indeed already on staging, but they were not present in my working tree after the staging merge — when I merged staging, my branch still contained an older ProductionListCard that did not reference these assets, and the merge resolution kept my older version. Only later, in a following commit where I aligned the list card with staging’s version, did I actually need those assets, so I pulled them in again from origin/staging — the files are byte-identical to staging (diff exit 0), so merging them does not create a conflict.

lucasodb:
Looks great now. I just want to make some minor changes to the grid card, just small spacing changes etc., I'll make a commit for that myself if you don't mind, that's a bit easier than explaining and via suggested changes.

lucasodb (approved):
All good to me now, if you agree with my last commit you can merge.

PR #448, opened 2026-05-14
change font of "archive" in navbar

lucasodb:
Chose a font that looks more like the viernulvier one.

before
image
after
image


lucasodb:
Oops started this from the wrong branch.

PR #453, opened 2026-05-14
use fallback thumbnails for productions that don't have media

### LINKED ISSUE #417: [DEV] Fallback thumbnail when no media is available
When a production has no media, a placeholder thumbnail should be displayed. Currently this is just a blank image, instead it should show something, like the VIERNULVIER (archive) logo.

lucasodb:
**Issue(s)**

Closes #417 

____

**Description**

Use default thumbnails with viernulvier logo as fallback when a production doesn't have media.

____

**Sources**


noahvsb (changes requested):
I like the dark mode svg, but the light mode svg feels a bit thinner and the "Archive" text is smaller and not well aligned. I think it's best to import the dark mode svg into inkscape and manually edit the colors for the lightmode version.

lucasodb:
> I like the dark mode svg, but the light mode svg feels a bit thinner and the "Archive" text is smaller and not well aligned. I think it's best to import the dark mode svg into inkscape and manually edit the colors for the lightmode version.

That's what I did, I used inkscape, and I also noticed that for some reason when I viewed the files in the ide itself the light move version looked different the way you described, but when I opened the files on their own they look the exact same, and also look the same on the website. Do you actually also see a difference on the website, or was this from looking at the files themselves in your ide?

noahvsb:
> > I like the dark mode svg, but the light mode svg feels a bit thinner and the "Archive" text is smaller and not well aligned. I think it's best to import the dark mode svg into inkscape and manually edit the colors for the lightmode version.
> 
> That's what I did, I used inkscape, and I also noticed that for some reason when I viewed the files in the ide itself the light move version looked different the way you described, but when I opened the files on their own they look the exact same, and also look the same on the website. Do you actually also see a difference on the website, or was this from looking at the files themselves in your ide?

I saw it on the website too.

noahvsb:
Maybe editing the svg directly might be easier, just copy paste the dark mode svg and edit the colours in the svg. That will ensure both stay the same shape.

lucasodb:
Should be good now...

noahvsb (approved):

PR #455, opened 2026-05-14
Feat/cms bulkedit

willemDC-Info:
**Issue(s)**

#122 solved

____

**Description**

Added bulk edit for productions is implemented. On the CMS side, the productions tab uses that path for regular inline edits, while tag editing gets special handling: normal tag edits preserve genre tags, and multi-row tag edits open a confirm step and then run a separate bulk flow that can either replace or diff the additional tags across all selected productions.
____

**Sources**


backend/src/routes/production/handlers/create.ts
trherber: I'm not sure but isn't there a lot of code duplication in this function? Wouldn't it work the same way like this?
```suggestion
export async function createProduction(
  server: FastifyInstance,
  request: FastifyRequest,
): Promise<ProductionWithBackwardsRefs | null> {
  const body = parseSchema(server, CreateProductionBodySchema, request.body);
  const { admin, current_time } = getMetadata(request);
  const tagIds = [...new Set(body.tags ?? [])];

  const fields: string[] = [];
  const placeholders: string[] = [];
  const values: unknown[] = [];
  let i = 1;

  const addField = (column: string, value: unknown) => {
    fields.push(column);
    placeholders.push(`$${i++}`);
    values.push(value);
  };

  for (const column of RequiredCreateColumns) {
    addField(column, getFieldValue(body, column));
  }
  for (const column of NullableCreateColumns) {
    addField(column, getNullableFieldValue(body, column));
  }

  fields.push("created_by", "updated_by", "created_at", "updated_at");
  placeholders.push(`$${i++}`, `$${i++}`, `$${i++}`, `$${i++}`);
  values.push(admin, admin, current_time, current_time);

  const client = await server.pg.connect();
  try {
    await client.query("BEGIN");

    const insertResult = await client.query<{ id: number }>(
      `INSERT INTO production (${fields.join(", ")})
       VALUES (${placeholders.join(", ")})
       RETURNING id`,
      values,
    );

    const row = insertResult.rows[0];
    if (!row) {
      await client.query("ROLLBACK");
      return null;
    }

    if (tagIds.length > 0) {
      await client.query(
        `INSERT INTO production_tag (production, tag, created_by, updated_by, created_at, updated_at)
         SELECT $1, t, $2, $2, $3, $3
         FROM UNNEST($4::int[]) AS t
         ON CONFLICT DO NOTHING`,
        [row.id, admin, current_time, tagIds],
      );
    }

    await client.query("COMMIT");
    return await getProductionById(server, row.id);
  } catch (err) {
    await client.query("ROLLBACK");
    server.log.error(err);
    throw err;
  } finally {
    client.release();
  }
}

```

frontend/src/components/admin/cms/productions/CmsProductionsTab.vue
```diff
@@ -588,6 +711,91 @@ async function saveTagEditorPanel(): Promise<void> {
...
+              existingAdditionalTagIds
+                .filter((tagId) => !removedTagIds.includes(tagId))
+                .concat(addedTagIds),
+            ),
+          );
+
+      await updateProduction(targetRow.id, {
+        tags: [...existingGenreTagIds, ...nextAdditionalTagIds],
+      });
+    }
```
trherber: Could this be parallelised with Promise.all instead of running sequentially with a for loop?
```suggestion
await Promise.all(
  targetRows.map((targetRow) => {
    const currentTagIds = extractProductionTagIds(targetRow.source);
    const existingGenreTagIds = currentTagIds.filter((tagId) => {
      const tag = tagsData.value.find((item) => item.id === tagId);
      return tag ? genreTypeIds.has(Number(tag.tag_type)) : false;
    });

    const existingAdditionalTagIds = currentTagIds.filter((tagId) => {
      const tag = tagsData.value.find((item) => item.id === tagId);
      return tag ? !genreTypeIds.has(Number(tag.tag_type)) : false;
    });

    const nextAdditionalTagIds =
      mode === "replace"
        ? desiredAdditionalTagIds
        : Array.from(
            new Set(
              existingAdditionalTagIds
                .filter((tagId) => !removedTagIds.includes(tagId))
                .concat(addedTagIds),
            ),
          );

    return updateProduction(targetRow.id, {
      tags: [...existingGenreTagIds, ...nextAdditionalTagIds],
    });
  }),
);
````

trherber (approved):
Great job! I made two minor comments below, feel free to take a look and see what you think but they are just small details. 

Aside from that, I noticed two small points about the user experience:

1. When bulk editing tags, the CMS shows the message: "If you close, only the current production will be edited." However, when I try it, closing doesn't actually edit the current production, it just cancels the entire action. I actually think canceling is the right behavior (if someone wants to edit just one production, they should just select one). But the current text is a bit confusing since it implies an edit will still happen.

2. I really love the option to choose between setting a fixed list of tags for all productions versus applying additions/removals across them. However, I feel like the distinction between "Replace" and "Add/Remove" isn't immediately intuitive at first glance because the three are listed directly underneath each other.

I asked AI to generate a quick concept of how this could look a bit more intuitive, which you can see in the screenshots below. If you think this is a good idea and you have the time, you can maybe still change this, but it is just a suggestion and personal taste. I definitely find it good enough already.

Current design:
Image

Suggestion:
Image

trherber (changes requested):
I approved in my previous review but the issue about the CMS saying it updates on cancel is maybe important to check out.

trherber (approved):
Looks great, thanks!!

PR #456, opened 2026-05-14
Related blogposts on production detail page

trherber:
**Issue(s)**

____

**Description**

Linked related blogposts on the production detail page

____

**Sources**


msth5:
When reviewing this, everything seemed fine originally, but then suddenly titles disappeared from the blogpost
image
And the detail page of said blogpost is empty now too.

edit:
Okay, found the culprit, the language setting changed in the middle out of nowhere to EN, shouldn't there be a default when current language isn't present? (NL -> EN -> FR)

msth5 (approved):
Looks good now, problems are solved

PR #462, opened 2026-05-15
fix tsconfig root dir

noahvsb:
**Issue(s)**

____

**Description**

The eslint plugin on vscode gave errors, I thought it was just a local issue, since eslint worked with the command. But today I decided to get to the cause of this and found an issue in the `eslint.config.js` regarding the tsconfig root dir.

There wasn't really an issue in the backend and shared folder, but I found the same inconsistency there, so I also changed it in those configs.

____

**Sources**

### DIFF
--- a/backend/eslint.config.js
+++ b/backend/eslint.config.js
@@ -1,10 +1,15 @@
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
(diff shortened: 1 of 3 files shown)


lucasodb (approved):
Looks good.

PR #463, opened 2026-05-15
Fix/scraper reworked endpoint

msth5:
**Issue(s)**

Continues on #389 

____

**Description**

With the endpoints being changed, the scraper need to reworked to use these endpoints and to delete the deprecated ones


backend/src/scraper/entities/event/event.ts
Acher0s, line 1: fix these warnings pls

backend/src/scraper/entities/production/production-tags.ts
Acher0s, line 1: fix warnings

pnpm-workspace.yaml
Acher0s, line 1: this seems like a security change, are you sure this is in scope for this PR?

Acher0s (changes requested):

pnpm-workspace.yaml
msth5: Got sneaked in while working at a  different pc I think, removed

backend/src/scraper/entities/production/production-tags.ts
Acher0s, line 1: warnings persist

Acher0s (changes requested):

Acher0s (approved):

Acher0s:
I'm assuming you actually did a test run of the scraper on a production build.

If not, do this so we're sure it works when merged with main.

```
docker compose up -d --build
docker exec <backend_container> node dist/backend/scripts/scrape.js
```

noahvsb:
I tested it and it works.

PR #440, opened 2026-05-14
add blogpost tab component

noahvsb:
**Issue(s)**

closes #439, #474
____

**Description**

Implements view, edit, create and delete on blogposts in CMS.

**TODO**

- [x] add pop-up for editing title and content, so all languages can be edited at the same time
- [x] add markdown editor for editing content
- [x] add markdown editor for creating content
- [x] edit associated productions

____

**Sources**

PR #468, opened 2026-05-16
Feat/add extra feats to tags tab

### LINKED ISSUE #472: [DEV] tags cms: add extra features
**Location**

/frontend

____

**Description**

- edit productions
- edit tag types
- edit name with editor panel

____

**Sources**

### LINKED ISSUE #475: [DEV] tags cms: show ID
**Location**

/frontend

____

**Description**

Show ID in tags cms.
____

**Sources**

noahvsb:
**Issue(s)**

closes #472, #475 

____

**Description**

Added feaures:
- [x] id is now displayed
- [x] productions can be edited with `CmsEditListPanel`.
- [x] tag types can be edited.
- [x] name can be edited using `CmsEditorPanel`.

____

**Sources**


noahvsb:
It's ready, however I'll wait with making it ready for review until #440 has been merged.

frontend/src/components/admin/cms/CmsEditListPanel.vue
```diff
@@ -56,7 +56,15 @@
...
               data-testid="edit-list-panel-tag"
             >
-              {{ id }}
+              <a
+                v-if="panel.urlBase"
+                :href="`${panel.urlBase}/${id}`"
+                class="cms-list-flair-link"
+                target="blank"
+              >
+                {{ id }}
```
Ikerstreamer, line 65: Wouldn't it be better to show the name of the production? As the id doesn't make it very easy to manage by a human. Likewise with the search box.

frontend/src/components/admin/cms/tags/CmsTagsTab.vue
```diff
@@ -227,6 +263,30 @@ async function persistTagPatch(
   }
 }
 
+function onCellClicked(event: CellClickedEvent<CmsTagGridRow>): void {
+  if (!event.data || !event.colDef.field) return;
```
Ikerstreamer, line 267: We should check if an editor panel is already open. Cause rn the behavior is inconsistent. Both editor panels will open and the one that is lowest in the html will be in top. Thus production will open over name but if u had production open and then open name it will look as if nothing happened until u close production adn then you'll see the name editor. Either make it so opening an editor closes the previous or make it so u must manually close the current editor before u can open a new one.

Ikerstreamer (commented):
My changes are for ui QoL but functionally the system works.

Ikerstreamer (changes requested):
If u can fix the 2 issues I think it is good to merge.

frontend/src/components/admin/cms/CmsEditListPanel.vue
noahvsb: Some tags are linked to thousands of productions, that requires thousands of fetches. I don't think it's worth it.

frontend/src/components/admin/cms/tags/CmsTagsTab.vue
Ikerstreamer, line 1: This works, and is probably the best solution.

Ikerstreamer (approved):
Nice UX improvement, everything else still works like before.

PR #479, opened 2026-05-16
Feat/cms media integration

willemDC-Info:
**Issue(s)**

solved #478 
____

**Description**

Adding media integration to the productions tab of the cms. Giving the user the ability to upload multiple images and video's when creating a production and the ability to edit, remove and preview all media of a production in the cms.

When uploading an image crops are automatically made and also added to the production.
____

**Sources**


lucasodb:
Make the crops generator actually make crops the same way viernulvier does. Same 5 crop types we use ourselves, and the corresponding resulotion/shape for each type.
- FE3_header: 1920x900
- FE3_home_featuredWide: 1920x600
- FE3_boxed: native shape, width is set at 1920
- nb_header: native shape, width is set at 500
- cms: native shape, width is set at 100

Simon-VdB:
image
I'm only able to edit the ones where there's already a url in the media column. I can't add a url to a production with no media yet.

Same for the images column. I'm only able to open the "Media" tab for productions that already have images.
Also I think it's a bit confusing that the two columns: media and images, open a pop up with the same name: "Media".

This is a pure functional review and once this is clearer, I will delve into your code.

frontend/src/services/cms/crop-generator.ts
Simon-VdB, line 1: contain mode distorts the image instead of letterboxing
The else branch for contain fills a white background but never adjusts dw/dh to preserve aspect ratio, so ctx.drawImage(img, 0, 0, sw, sh, 0, 0, dw, dh) stretches the source to exactly crop.width × crop.height. The letterbox comment is wrong — the image is stretched.
contain isn't used in the current CMS_CROP_DEFINITIONS (only cover and native are), so this doesn't break anything today, but the moment someone adds a contain crop it will silently distort.

frontend/src/services/cms/grid.ts
Simon-VdB: Since the bulk-edit response doesn't include images, clearing them is technically defensible, but in practice it means every time you edit a title or tag the image thumbnails disappear from the grid until the page is reloaded. A more graceful approach would be to keep the existing image state.

Simon-VdB (changes requested):
Looks good to me for the most part. Some minor changes

Ikerstreamer (commented):
Your tests look solid and the changes u made make sense. I didn't look at all of Willems work yet.

Simon-VdB (approved):
Let's merge this.

PR #482, opened 2026-05-17
Featured blog post on landing page

trherber:
**Issue(s)**

____

**Description**

Adds a featured blog post on the landing page. The featured post is currently selected through a hardcoded post ID (`FEATURED_POST_ID`). An alternative approach would be to automatically display the latest blog post but that would remove the ability to pin a specific post as featured. The cover image is the first image in the markdown content since blog posts do not have media stored in the db.
____

**Sources**


msth5:
If the bogpost is very short, the first letter capitalized and big, conflicts with the read more/lees meer link "below" it (it should be below, but in this scenario it isn't)

msth5 (approved):

PR #487, opened 2026-05-17
create create-default-blog script

### LINKED ISSUE #486: [DEV] create a create-blog.ts script to insert a default blog (with id = 1)
**Location**

`backend/scripts`

____

**Description**

create a create-blog.ts script to insert a default blog (with id = 1)

____

**Sources**

noahvsb:
**Issue(s)**

#486 
____

**Description**

image

____

**Sources**

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -35,6 +35,7 @@
     "lint:fix": "eslint . --fix",
     "check": "pnpm run build && pnpm run coverage && pnpm run lint",
     "create-admin": "tsx scripts/create-admin.ts",
+    "create-default-blog": "tsx scripts/create-default-blog.ts",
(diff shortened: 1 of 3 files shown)


lucasodb:
Maybe it would be useful to also add a script that creates a default blog post? This could create a very basic production then too which it refers.

noahvsb:
> Maybe it would be useful to also add a script that creates a default blog post? This could create a very basic production then too which it refers.

I don't really see the point. You can use the CMS to make this. The reason for this script is that you can't use the CMS to create a blog. I also wouldn't have too many of these scripts because they don't have zod validation and are dangerous.

msth5 (approved):
Might even be a bit over the top for what it's used for

Issue #476, opened 2026-05-16
[DEV] productions cms: show ID


noahvsb:
**Location**
frontend

____

**Description**

show ID in productions cms, the IDs need a hyperlink to the public page of that specific production
____

**Sources**

Issue #450, opened 2026-05-14
[BUG] doesn't refresh grid automatically on language change in CMS


noahvsb:
**Description**

I edited a description and added a french translation when my language was set to NL.
Then I set my language to FR to see the new translation and I saw it wasn't updated.
I had to manually refresh in order to update it. 

____

**Environment**

brave browser on linux

**Sources**

noahvsb:
Seems to be fixed on all tabs, except the productions tab.

PR #488, opened 2026-05-17
Adding production id to the production cms with hyperlink to detail page

willemDC-Info:
**Issue(s)**

solves:

#476 
#450 
____

**Description**

Adding production id's to the cms to link to the production details page for ease of use.
Also fixing some small bugs in production cms when updating production in a different language then your currently selected one.
____

**Sources**


noahvsb:
@willemDC-Info you made it a bit too complicated, you can add the link with two lines, see #489.

noahvsb (changes requested):
Great job using my implementation as I asked, but turns out my implementation was wrong. The current language should be included in the URL, see #489.

noahvsb (approved):

PR #489, opened 2026-05-17
make id in blogposts cms link to the public page

### LINKED ISSUE #484: [DEV] blogpost CMS: id column needs a hyperlink to the public page
**Location**

frontend -> cms -> blogpost

____

**Description**

The id's in the id column should link to the public page: e.g. `/blog/post/:id`.
____

**Sources**

noahvsb:
**Issue(s)**

#484 
____

**Description**

Did the issue and also added a test file for the composable.

____

**Sources**

### DIFF
--- a/frontend/src/composables/useCmsBlogPostGrid.ts
+++ b/frontend/src/composables/useCmsBlogPostGrid.ts
@@ -67,6 +67,8 @@ export function useCmsBlogPostGrid(options: {
       field: "id",
       minWidth: 50,
       maxWidth: 100,
+      cellRenderer: (params: { value: number }) =>
+        `<a href="/blog/post/${params.value}" class="text-ink-primary underline hover:text-ink-secondary transition-colors">${params.value}</a>`,
     },
     {
       headerName: options.t("cms.columns.title"),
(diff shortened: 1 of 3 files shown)


frontend/src/composables/useCmsBlogPostGrid.ts
```diff
@@ -67,6 +67,8 @@ export function useCmsBlogPostGrid(options: {
       field: "id",
       minWidth: 50,
       maxWidth: 100,
+      cellRenderer: (params: { value: number }) =>
+        `<a href="/blog/post/${params.value}" class="text-ink-primary underline hover:text-ink-secondary transition-colors">${params.value}</a>`,
```
Simon-VdB: the link is missing the language prefix and will not resolve.

Simon-VdB: A raw <a href> triggers a full page reload instead of a router transition. Two reasonable fixes:

Resolve the URL via router.resolve({ name: RouteNames.BLOG_POST_DETAIL, params: { lang, id } }).href and keep using <a> (still full reload, but at least correct).
Better: use AG Grid's custom-cell-renderer component pattern with <RouterLink> (similar to how useCmsProductionGrid.ts uses a component-style renderer for media.

Simon-VdB (changes requested):
Some blocking issues, sorry

frontend/src/composables/useCmsBlogPostGrid.ts
noahvsb: With the target being _blank, I'm guessing it triggers a full page load anyway. So I won't fix this.

Simon-VdB (approved):
Great thank you.

PR #492, opened 2026-05-18
normalize production detail HTML and trim excessive breaks

lucasodb:
**Issue(s)**

____

**Description**

Tightened how production detail text is normalized and how it is styled. In `parsers.ts`, parseAndSanitizeContent now caps long runs of literal newlines, collapses excessive \<br> chains, strips trailing \<br> before \</p> (so breaks don’t stack with block margins), minifies newline-only whitespace between HTML tags, and converts plain (non-HTML) copy into safe \<p> / \<br /> HTML before sanitizing. 
In `DetailsSection.vue`, rich blocks use a shared layout: wrapper white-space: normal (so inter-tag newlines don’t fake extra gaps), pre-line only on p/li for in-flow line breaks, explicit margins for paragraphs, lists, and \<hr> (Preflight had removed default spacing outside .prose), plus description_2 aligned with the same pattern as the main body. Together this removes exaggerated empty bands from CMS noise while keeping intentional paragraph and section rhythm.

Some examples:

Before:
image
image
After:
image
image

____

**Sources**


noahvsb (approved):
Nice, the detail page looks way cleaner now 👍

PR #493, opened 2026-05-18
append viernulvier url to cms_files media paths that don't have it

lucasodb:
**Issue(s)**

____

**Description**

In the description and info fields, there are often sponsor logo's and other images shown, which are just a viernulvier url, that looks like  `https://www.viernulvier.gent/cms_files/Image/...`. Often though, the viernulvier prefix is left out, and the paths are just `/cms_files/Image/...`, which of course still works on the viernulvier website, but not on ours. 
The html for the detail fields is now first scanned to see if there are local image paths, and if so, `https://www.viernulvier.gent` gets appended, so the image still gets loaded. 

____

**Sources**


Acher0s (dismissed):

noahvsb:
@lucasodb merge conflicts.

noahvsb (approved):
Nice improvement.

PR #494, opened 2026-05-18
add codecov.yml config to fix issue + general improvements

noahvsb:
**Issue(s)**

Validation of #493 won't fail now

____

**Description**

- Made codecov not do the validation.
- Added flags for frontend and backend so the coverage is split up between both

____

**Sources**

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,0 +1,15 @@
+coverage:
+  status:
+    project: off
+    patch: off
+
+comment: false
+
+flags:
+  frontend:
+    paths:
+      - frontend/
+
+  backend:
+    paths:
+      - backend/
--- a/.github/workflows/pr-staging.yml
+++ b/.github/workflows/pr-staging.yml
@@ -59,6 +59,8 @@ jobs:
         uses: codecov/codecov-action@v5
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
+          flags: ${{ matrix.project }}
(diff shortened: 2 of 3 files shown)


noahvsb:
image

.github/codecov.yml
Ikerstreamer, line 1: When I run 
```
curl --data-binary .github/codecov.yml https://codecov.io/validate
```
I do not get a 200 response code. In codecovs docs it says this url should validate the yml file.

https://docs.codecov.com/docs/codecov-yaml

Ikerstreamer: Never user error.

```
curl -X POST --data-binary @.github/codecov.yml https://codecov.io/validate
Valid!

{
  "coverage": {
    "status": {
      "project": false,
      "patch": false
    }
  },
  "comment": false,
  "flags": {
    "frontend": {
      "paths": [
        "^frontend/.*"
      ]
    },
    "backend": {
      "paths": [
        "^backend/.*"
      ]
    }
  }
}
```

It seems to validate succesfully.

Ikerstreamer (approved):
On the [site](https://app.codecov.io/gh/SELab-2/viernulvier-1/flags/fix%2Fnavbar-overlap-on-half-screen) it states that flag analytics is disabled which sounds problematic since we are splitting up the analysis by flags.

Im gonna mark this as approaved as it seems the code changes are sound and it might be a config issue on the site.

PR #495, opened 2026-05-18
Fixed navbar overlap on half screen

### LINKED ISSUE #491: [BUG]
**Description**

Nav bar of the cms does not switch to hamburger menu in time since it has more text in the navbar in comparison to public part of the website.

____

**Environment**

I'm having this issue on brave when sizing down the screen.

**Sources**

Image

noahvsb:
**Issue(s)**

#491

____

**Description**

Fixed #491 by changing the requirements for the hamburger to appear from `sm` (640 px) to `lg` (1024 px).
The overlap would also be fixed with around 880 px, but I found the standardized 1024 px in the `DESIGN.md` and it gives us a bit more buffer when adding new links in the navbar.
____

**Sources**

### DIFF
--- a/frontend/src/assets/stylesheets/navbar.css
+++ b/frontend/src/assets/stylesheets/navbar.css
@@ -20,12 +20,12 @@
 
 /* ===== NAV LEFT ===== */
 .nav-left { @apply flex items-center; }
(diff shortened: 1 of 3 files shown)


willemDC-Info (approved):
Small and simple fix. Looks nice on all pages so this can be merged to staging.

PR #497, opened 2026-05-18
Small fixes blogpost detail

trherber:
…lback

**Issue(s)**

____

**Description**

The blog post detail page now uses the notFound component for consistent not found states across the application. Additionally, when a language map contained an empty string, that empty value was previously displayed. It now falls back to the first available non-empty translation instead.
____

**Sources**


lucasodb (approved):
Good job.

PR #499, opened 2026-05-18
loading states

trherber:
**Issue(s)**

____

**Description**

I created pulsing placeholders (skeleton states) for the production cards that are shown while the productions are loading.

The skeleton state is only used on the initial page load. For filter changes skeletons caused too much flicker so I switched to a lighter loading state where existing productions are slightly faded to indicate new data is loading without a disruptive visual change.

This is just an enhancement idea, if it’s not an improvement we can leave it out. I do think it’s useful though that the productions become slightly transparent while loading, so there’s a clear visual cue that the page is updating and not stuck.

Note:
I already had an open PR for this, but because of merge conflicts I started over from staging.

____

**Sources**


lucasodb (approved):
Good job, you adapted the skeletons well to the current layout, this is good feedback for users that the page is loading.

Issue #503, opened 2026-05-18
[DEV] Make the title, supertitle and artist colums use the md editor.


Ikerstreamer:
**Location**

In `CmsProductionTab.vue` we need to change how these columns are handeld.
____

**Description**

The columns should use a md editor when editing just like the descriptions.

PR #504, opened 2026-05-18
Feat/cms production markdown editors

Ikerstreamer:
**Issue(s)**

#503

____

**Description**

Make the fields Title, Supertitle and artist all be md fields with all 3 possible languages.


frontend/src/components/admin/cms/productions/CmsProductionsTab.vue
```diff
@@ -897,6 +900,7 @@ function closeBulkEditConfirm(): void {
 }
 
 async function confirmBulkEdit(): Promise<void> {
+  console.log(pendingBulkEditAction);
```
Simon-VdB: This shouldn't be in production code...

noahvsb (changes requested):
Set `editable: false` in the useCmsProductionsGrid.ts composable

Simon-VdB (changes requested):
Image

Whenever I double click a cell to do inline editing, it also opens the side panel (on the first click). This is really annoying. I would like to be able to choose between inline edit: double click, OR side panel edit: single click. But NOT both.
Also the inline edit is now broken: when i edit a cell and hit enter, it shows the new version, but when I refresh, it goes back to the old version, indicating that it's not being saved. 

Also the bulk edit doesn't work anymore. I select 2 rows, I do single click on artist cel. I change somthing, I click save, I get a bulk edit confirmation modal, but then when I click continue, it doesn't save the changes, nothing happens to the db, even though I get a green confirmation message. This is maybe due to an issue in `applyTransactionAsync`?

Al

noahvsb:
> Whenever I double click a cell to do inline editing, it also opens the side panel (on the first click). This is really annoying. I would like to be able to choose between inline edit: double click, OR side panel edit: single click. But NOT both. Also the inline edit is now broken: when i edit a cell and hit enter, it shows the new version, but when I refresh, it goes back to the old version, indicating that it's not being saved.

This has been discussed in the discord, it's a good idea to have single click do the inline editing and double click do the popup. But for consistency we would also have to change it for the other tabs, which we don't have time for. Just setting `editable: false` is the fix here.

Simon-VdB (approved):

noahvsb (approved):

PR #505, opened 2026-05-19
don't show empty banner when production has no media

### LINKED ISSUE #418: [DEV] Don't display banner when there is no media
When a production has no media, a default blank banner is displayed. This doesn't look good, especially now that we use the cards to display production info, so a better solution should be found for this.

lucasodb:
**Issue(s)**

Closes #418 

____

**Description**

Before:
image

After:
image

____

**Sources**


noahvsb (approved):
Looks good and code is clean.

PR #506, opened 2026-05-19
add url query for the selected view

lucasodb:
**Issue(s)**

____

**Description**

There is now a `?view=` query added to the url, which saves the layout of the productions page. List is still the default, so only for grid will `?view=grid` be visible in the url. 

Another completely unrelated change, but a bit dumb to make a pr for this; added padding above the first row of productions cards in grid view, because they otherwise sit too close to the collapsed genre row (list cards sat at same distance, but these have padding on top within the cards themselves, which makes it look like there is more space between).

____

**Sources**


willemDC-Info (approved):
It is a nice and small feature and fix. All looks well and seems to work well on different window sizes.
Well Done!!

PR #509, opened 2026-05-19
Fix dark mode contrast on CMS buttons + Cancel label on bulk-edit modal

Simon-VdB:
**Issue(s)**

Closes #507
Closes #508

____

**Description**

Two small UI fixes in the CMS:

1. **Dark mode contrast** (#507) — Primary CMS buttons (`Yes, update`, `Add`, `Save`, language pills, ...) used `bg-surface-inv text-ink-on-inv`. In dark mode `--surface-inv` resolves to `#1C1A17` which is **identical to** `--surface-0` (the page background), so the buttons were invisible. Swapped to `bg-accent-dark text-surface-0 hover:bg-accent-dark-hover` — the same pattern already used in `ChangePasswordModal.vue`. The `--accent-dark` token swaps properly per theme (dark in light mode, light in dark mode), giving high contrast in both.

   Touched classes: `cms-side-save`, `cms-add-button`, `cms-list-add`, `cms-form-submit`, `cms-language-pill.active` — across `cms-view.css`, `cms-create-modal.css`, and the four components with per-component duplicates of `.cms-side-save`/`.cms-language-pill.active`.

2. **Duplicate Close labels** (#508) — The bulk-edit confirm modal had two buttons labelled "Close" (header X and footer dismiss). The footer button now uses the existing `general.cancel` key → "Cancel" / "Annuleer" / "Annuler".

____

**Verification**

- `npm run lint` → 0 errors (only pre-existing v-html warnings)
- `CmsProductionsTab.test.ts` → 38/38 passing
- Dark mode computed styles checked in the browser preview: all touched button classes now render `background #EDE8DF` on page bg `#1C1A17` — full contrast.
- Light mode unchanged: `--accent-dark` is `#2B2826` in light mode, same as the old `--surface-inv`, so the look is identical.

Larger inverted-surface sections (footer, navbar, hero) keep `bg-surface-inv` — that's a separate design discussion out of scope.

PR #513, opened 2026-05-19
Fix bulk-edit cell staleness in CMS productions grid

Simon-VdB:
**Issue(s)**

Addresses the bulk-edit staleness flagged by @willemDC-Info in [#509](https://github.com/SELab-2/viernulvier-1/pull/509) and by @Simon-VdB in [#504](https://github.com/SELab-2/viernulvier-1/pull/504) (\"green confirmation message but DB / grid not updated\"). Same underlying bug from three angles.

____

**Root cause**

`persistBulkProductionPatch` on staging had three problems stacked on top of each other:

1. **Duplicate API call** — `bulkUpdateProductions` is called once, then `loadCmsData()` runs, then it's called again. The first call's result is discarded; clear merge artifact.
2. **`loadCmsData()` orphans the `targetRows` refs** — `loadCmsData` calls `rebuildRows`, which replaces every row in `rowData.value` with a freshly built object. The `targetRows` parameter that was handed in earlier now points at row objects that are no longer in the grid. The subsequent `applyUpdatedProductionToRow(row, …)` therefore mutates dead objects.
3. **No re-render hint to AG Grid** — even with the orphaning fixed, AG Grid needs to be told to redraw the cells. Without `getRowId`, the client-side row model can only match nodes by reference identity, so any `applyTransactionAsync({ update })` or `refreshCells` strategy silently no-ops.

Net effect users saw: bulk save returns success and DB is updated, but cells keep showing the pre-save value until a full page reload.

____

**Fix**

- Drop the duplicate `bulkUpdateProductions` call and the `loadCmsData()` reload in `persistBulkProductionPatch`.
- Mutate each target row in place, then call `gridApi.refreshCells({ rowNodes, force: true })` on the affected nodes. `force: true` bypasses the value cache so AG Grid actually redraws.
- Add `getRowId: ({ data }) => String(data.id)` to the grid so `getRowNode(id)` returns the right node — required for the refresh to find anything.

____

**Tests**

- Restored the `expect(api.bulkEditConfirmOpen.value).toBe(false)` assertion that was dropped in #504 (which was a band-aid for this same bug).
- Added a regression test asserting:
  - `bulkUpdateProductions` is called **exactly once** (catches the duplicate-call regression coming back)
  - `refreshCells` is called **once** with the right `rowNodes` and `force: true`

Result: 69/69 passing locally. `vue-tsc` and `eslint` clean on the touched files.

____

**Verification**

I could not exercise the CMS path in the browser preview (admin login requires backend; backend wasn't reachable in my sandbox), but the unit test directly asserts the `refreshCells` contract that fixes the reported symptom. Worth a manual check after merge:

1. Select ≥2 productions, click an Artist/Title/Supertitle/Teaser cell, edit, save → confirm modal → \"Yes, update\".
2. Cells should immediately show the new values without a page reload.
3. Network tab should show **one** PATCH bulk-update request, not two.

### DIFF
--- a/frontend/test/components/admin/cms/productions/CmsProductionsTab.test.ts
+++ b/frontend/test/components/admin/cms/productions/CmsProductionsTab.test.ts
@@ -502,6 +502,47 @@ describe("CmsProductionsTab", () => {
     expect(api.bulkEditConfirmOpen.value).toBe(false);
   });
 
+  it("refreshes affected grid cells exactly once after a successful bulk patch", async () => {
+    const wrapper = await mountTab();
+    const api = (wrapper.vm as any).$?.exposed.__test;
+    const row = api.rowData.value[0];
+    const secondRow = {
+      ...row,
+      id: row.id + 1,
+      source: { ...row.source, id: row.id + 1 },
+    };
+
+    const fakeNodeA = { id: String(row.id) };
+    const fakeNodeB = { id: String(secondRow.id) };
+    const getRowNode = vi.fn((id: string) => (id === String(row.id) ? fakeNodeA : fakeNodeB));
+    const refreshCells = vi.fn();
+
+    api.gridApi.value = {
+      getSelectedRows: () => [row, secondRow],
+      getRowNode,
+      refreshCells,
+    };
+
+    vi.spyOn(productionsService, "bulkUpdateProductions").mockResolvedValueOnce([
+      { ...mockProduction, id: row.id },
+      { ...mockProduction, id: secondRow.id },
+    ]);
+
+    api.onCellClicked({
+      data: row,
+      colDef: { field: "descriptionOne", headerName: "Description" },
+    });
+    await api.saveEditorPanel();
+    await api.confirmBulkEdit();
+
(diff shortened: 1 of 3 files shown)


frontend/test/components/admin/cms/productions/CmsProductionsTab.test.ts
```diff
@@ -502,6 +502,47 @@ describe("CmsProductionsTab", () => {
     expect(api.bulkEditConfirmOpen.value).toBe(false);
   });
 
+  it("refreshes affected grid cells exactly once after a successful bulk patch", async () => {
```
willemDC-Info, line 505: It would appear the fix is not fully done yet. If I edit in dutch and would switch the localisation, the cells appear empty again. If I then change back to dutch the cells are stale again.

Also the cell still has its old makeup as the ag-grid still conciders it empty at this point

Image

willemDC-Info (changes requested):

willemDC-Info (approved):
Lovely changes. All looks well and also works nicely and spotless.
An amazing job Simon!!!

PR #518, opened 2026-05-19
"select all" checkbox now selects on active filters

willemDC-Info:
**Issue(s)**

____

**Description**

Fixing a bug where when selecting all productions in the cms it selected all prodution regardless of the current filters that had been set up causing the user to possibly edit more productions then wanted

____

**Sources**


noahvsb:
Only 1 line necessary, hallelujah :pray:

Simon-VdB (approved):
It works, very well done.

PR #519, opened 2026-05-19
show tags for all tag types in the filter panel

lucasodb:
**Issue(s)**

____

**Description**

The filter panel only showed all tags with tag types 'genre' and 'tag'. Now it shows all tags, grouped by each tag type. You can test this by adding a tag type in the cms, then adding a tag for this tag type, and assigning this tag to a production. Then the row for this tag type should show up in the filter panel. The first commit is mostly just moving code from the productionsview to a shared separate component, no real logic changes.

The idea is to allow the creation of more tag types in the future, which can be used to define types of series of productions.

image
____

**Sources**


frontend/src/views/ProductionsView.vue
```diff
@@ -617,8 +546,49 @@ const MAX_SEARCH_TERMS = 20;
...
+
+/**
+ * Pre-bound `useFittingPills` rows for the expandable filter panel. They live here (not under
+ * `v-if` children) so their lifecycle aligns with ProductionsView
+ */
+type FilterPanelFitTag = { id: number; label: string };
+
+/** Pre-allocated fitting slots (worst-case section counts). Clamp if CMS ever exceeds. */
+const PANEL_FIT_GENRE_POOL = 6;
+const PANEL_FIT_ACCENT_POOL = 14;
```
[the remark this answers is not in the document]
lucasodb: @copilot apply changes based on [this feedback](https://github.com/SELab-2/viernulvier-1/pull/519#discussion_r3266279717)

frontend/src/composables/useFittingPills.ts
```diff
@@ -19,6 +19,11 @@ interface UseFittingPillsOptions {
   trailingControlGapPx?: number;
   /** Used only when widths are not measurable (e.g. JSDOM). */
   fallbackVisibleCount?: number;
+  /**
+   * When false, this instance does not add a `window.resize` listener. Use when multiple
+   * rows share one resize handler from a parent view (Productions archive panel).
+   */
+  attachWindowResizeListener?: boolean;
```
noahvsb, line 30: It's better to just make this default to true. Then you don't need to mess with undefined behaviour in the following few changes.

noahvsb (changes requested):
A ton of errors in the tests. I have no idea why they are still succeeding.

Image

I suggest also adding a few more tests. But they don't have to be very complicated.

frontend/src/views/ProductionsView.vue
Simon-VdB, line 1: createPanelFitSlot() is called from inside the watch(panelFilterSections, …, { immediate: true }) callback, but tagsById / tagTypesById are populated asynchronously inside onMounted (line 1548-1549). So the first time a new tag type appears, the watcher runs after setup. createPanelFitSlot → useFittingPills then calls onMounted / onUnmounted / watch outside any active component instance:

Vue logs the dev-mode warning onMounted is called when there is no active component instance.
The internal watch(items), watch(trailingControlEl), watch(rowEl) in useFittingPills are created with no owning scope, so they never get torn down when the component unmounts → memory leak on every CMS-side tag type addition while the page is open.
The ResizeObserver setup inside onMounted simply never runs for those slots; resize still works by coincidence via the shared productionsFitPillsOnWindowResize handler, but observation of rowEl's own size changes (e.g. parent layout reflow without a window event) is silently lost.
Fix: wrap each slot in an effectScope, store the scope alongside the slot, and stop it when you delete the slot:
```
import { effectScope, type EffectScope } from "vue";

function createPanelFitSlot(fallbackCap: number) {
  const scope = effectScope(true);
  const slot = scope.run(() => {
    const candidatesRef = ref<FilterPanelFitTag[]>([]);
    const fitting = useFittingPills(candidatesRef, { … });
    return { candidatesRef, ...fitting };
  })!;
  return { ...slot, _scope: scope };
}
// on delete: slot._scope.stop();
```
You'll also want a final onScopeStop/onUnmounted that stops every remaining scope. (The effectScope route also fixes the lifecycle-hook warnings because useFittingPills's onMounted falls back to running its callback eagerly when there's no instance — but only inside a scope is the watcher cleanup correct.)

Simon-VdB (changes requested):

Simon-VdB (approved):
Looks good, let's merge

noahvsb (approved):

PR #520, opened 2026-05-19
add metadata description and open graph

noahvsb:
**Issue(s)**

____

**Description**

Added metadata description and open graph. These are used to generate embeds when you paste the URL somewhere. Always nice to have.

____

**Sources**


lucasodb (approved):
Looks good.

PR #532, opened 2026-05-19
fix event loading text

noahvsb:
**Issue(s)**

____

**Description**

fix event loading text

it was previously set to the saving text

____

**Sources**


lucasodb (approved):
Good fix.

PR #541, opened 2026-05-19
fix(api): forward FormData as multipart so CMS image upload actually persists

### LINKED ISSUE #533: [BUG] Can't add image in CMS
**Description**

When creating a new production, I add an image, but it doesn't get actually added.
When later trying to add it again, same problem (To any exiting production, not only self added)

____

**Environment**

**Sources**

Simon-VdB:
Closes #533

## Root cause

`apiFetch` in [frontend/src/services/api.ts](frontend/src/services/api.ts) called `JSON.stringify` on every body and set `Content-Type: application/json` whenever a body was present. The CMS media upload in [frontend/src/services/cms/media-upload.ts](frontend/src/services/cms/media-upload.ts) passes a `FormData` (image blob + crop blobs) to `apiFetch`. `JSON.stringify(formData)` returns `\"{}\"`, so the backend handler at [backend/src/routes/media/handlers/create.ts](backend/src/routes/media/handlers/create.ts) — which requires `multipart/form-data` — received an empty JSON body and rejected the request.

The error was caught and only `console.error`'d in `submitCreateProduction` (CmsProductionsTab.vue:1417-1423), and rendered as a non-blocking toast in `onMediaPreviewImageSelected`, so users saw no clear feedback — matching the issue: \"I add an image, but it doesn't get actually added\".

The existing media-upload tests didn't catch this because they mock `apiFetch`, never exercising the real fetch path.

## Fix

Detect native `BodyInit` values (`FormData`, `Blob`, `URLSearchParams`, `ArrayBuffer`, `ArrayBufferView`, `ReadableStream`) in `apiFetch` and forward them to `fetch` unchanged. Skip the JSON `Content-Type` header for these so the browser sets the correct header (including the multipart boundary).

## Test plan

- [x] `npx vitest run test/services/api.test.ts` — new FormData/Blob/URLSearchParams test cases all green
- [x] `npx vitest run` — full suite 1577/1577 ✅
- [x] `npm run lint` ✅
- [x] `npx vue-tsc --noEmit` ✅
- [ ] Manueel met draaiende backend: in CMS productions tab een nieuwe productie aanmaken mét image → image verschijnt in de image kolom na opslaan
- [ ] Manueel: op image kolom van een bestaande productie klikken → \"Add image\" → image wordt zichtbaar in de gallery

### DIFF
--- a/backend/test/routes/auth/logout.test.ts
+++ b/backend/test/routes/auth/logout.test.ts
@@ -63,23 +63,32 @@ describe("Logout", () => {
   });
 
   test("POST /api/v1/auth/logout — evicts jti from denylist after token expiry", async () => {
-    const jti = server.generateJti();
-    const shortLivedCookie = server.jwt.sign(
-      { id: 404, username: "Karel", jti },
-      { expiresIn: "1s" },
-    );
+    // Use fake timers so the assertion below isn't racing against the
+    // setTimeout that the logout handler schedules. On slow CI the inject
+    // call could yield to the event loop long enough for a sub-second
+    // timer to fire before the assertion runs.
+    vi.useFakeTimers({ shouldAdvanceTime: false });
+    try {
+      const jti = server.generateJti();
+      const shortLivedCookie = server.jwt.sign(
+        { id: 404, username: "Karel", jti },
+        { expiresIn: "1s" },
+      );
 
-    await server.inject({
-      method: "POST",
-      url: "/api/v1/auth/logout",
-      cookies: { session: shortLivedCookie },
-    });
+      await server.inject({
+        method: "POST",
+        url: "/api/v1/auth/logout",
+        cookies: { session: shortLivedCookie },
+      });
 
-    expect(server.tokenDenylist.has(jti)).toBe(true);
+      expect(server.tokenDenylist.has(jti)).toBe(true);
 
-    await new Promise((resolve) => setTimeout(resolve, 1500));
+      vi.advanceTimersByTime(1500);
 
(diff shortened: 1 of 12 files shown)


Simon-VdB:
Added 3 follow-up fixes in commit a3e8077f after manual testing surfaced more issues in the same flow:

**1. Media preview modal too tall (footer clipped on short viewports)**  
`.cms-media-preview-body` had `min-height: 500px`. With `h-[85vh]` on the modal and `overflow-hidden`, the footer (Close / Delete / Add image) was pushed past the modal on smaller windows. Dropped the fixed min-height and clamped the min-width.

**2. HEIC images failed silently**  
HEIC can't be decoded into an ``, so `generateAllCrops` rejected. The error went into `saveError`, but the media preview modal didn't render it. Now:
- File input `accept` narrowed to `image/jpeg,image/png,image/webp,image/gif` (media preview modal + create production modal)
- JS MIME validation as safety net
- `saveError` is now displayed inside `CmsMediaPreviewModal`
- New i18n key `cms.errors.unsupportedImageType` (nl/en/fr)

**3. Gallery showed only the newly uploaded image**  
`loadCmsData()` fetches images via a fire-and-forget IIFE, so after `await loadCmsData()` `imagesByProductionId` is still stale. The upload handler reopened a single-image preview, losing the other images until the user closed and reopened the modal. Now refetches images for the production directly and reopens `openImageGalleryPreview` with the full list, focused on the new image.

All 1577 tests still pass. Could not verify end-to-end visually in my preview (backend Bad Gateway here).

msth5 (approved):
Real nice job

